PR #10, opened 2019-01-29
small update to readme

beav:


README.md
```diff
@@ -2,3 +2,24 @@
...
+
+# how to run with local drift-backend
+
+make a js file somewhere that looks like this
+
+```
+module.exports.routes = {
+           '/r/insights/platform/drift' : { host: 'http://localhost:8080' },
+}
+```
```
ehelms: Let's put this into the repository since it should be consistent for all users of drift-frontend to use at some point?

ehelms: Thoughts here @beav  ?

beav: I put in a PR to store it in drift-backend at https://github.com/RedHatInsights/drift-backend/pull/20/files. Once that's merged, I'll update this PR to reference that file.

ehelms:
Can be updated now.

PR #12, opened 2019-01-30
Allow passing hosts params to UI and down to API

ehelms:


ehelms:
@beav This puts a burden on the backend so we should agree on the param name and structure. in testing this, here is the request made to the backend:

```
10.131.29.163 - - [30/Jan/2019 01:36:58] "GET /r/insights/platform/drift/v0/compare?hosts[]=1&hosts[]=2&hosts[]=3 HTTP/1.1" 200 -
--

**Comment by beav** [comment]
> @beav This puts a burden on the backend so we should agree on the param name and structure. in testing this, here is the request made to the backend:
> 
> ```
> 10.131.29.163 - - [30/Jan/2019 01:36:58] "GET /r/insights/platform/drift/v0/compare?hosts[]=1&hosts[]=2&hosts[]=3 HTTP/1.1" 200 -
> ```

I can support that format just fine. I think if we get a very large number of hosts, we'll hit other issues before URL length :smile_cat: 

It's up to you if you want to commit this now, or wait until I support it on the backend first.

**Comment by ehelms** [comment]
@beav The other thought that occurred to me was if this should be a POST with a JSON payload enumerating hosts. Does OpenAPI spec indicate any direction?

**Comment by beav** [comment]
Could you do `compare?hosts=1,2,3,4`? This seems like the cleanest way from what I read (personal opinion)

**Comment by beav** [comment]
> I can go that route with some custom parsing code on our side. The query string parses I have encountered don't recognize this as an 'array' syntax as the typical is to do either (at least on the UI side, are you seeing different on Python?):
> 
> ```
> host=1&host=2
> ```
> 
> or
> 
> ```
> host[]=1&host[]=2
> ```

ya, on my side it didn't recognize either as an array. I thought maybe there was some flask magic to auto-create the array but I didn't see any.

The `1,2,3` lets me just grab the string and run `split(',')` on it to create the array.

**Comment by ehelms** [comment]
Rebased

**Comment by beav** [comment]
is it easy to change to `host_ids[]=1&host_ids[]=2`? I can go with `hosts[]=1&hosts[]=2` but adding `ids` is more descriptive

**Comment by ehelms** [comment]
Now:

```

10.129.33.184 - - [31/Jan/2019 17:00:42] "GET /r/insights/platform/drift/v0/compare?host_ids[]=1&host_ids[]=2&host_ids[]=3 HTTP/1.1" 200 -
```

beav:
can't get much better than 2 acks on a 3 person team. merging

PR #13, opened 2019-01-30
Implement 'Add System'

johnsonm325:


beav:
I'm not a JS expert but it seems reasonable to me based on the screenshot

add-system.scss
```diff
@@ -0,0 +1,2 @@
+// js generates class '.page__{rootClass}' that wraps the content in your page added to the root tag
+// Found in Routes.js as the rootClass
```
beav: does this file need to exist? its all comments

johnsonm325: I don't think it does right now, but I think it will eventually. I can delete for now though. No need to keep

src/SmartComponents/DriftTable/AddSystem.js
```diff
@@ -0,0 +1,23 @@
...
+export class AddSystem extends Component {
+    constructor(props) {
+        super(props);
+    }
+
+    render() {
+        return (
+            <Button
+                variant='primary'
+                onClick={ this.props.getAddSysModal }>
```
ehelms: This is a bit nitpicky, but I tend to prefer consistency when it comes to shortening names. I find this helps with developer predictability and readability. In this case, since `System` is used in most places I would prefer `getAddSystemModal`.

PR #14, opened 2019-01-31
Add redux setup for actions and reducers

ehelms:
This is a lot of changes, as implementing the full redux workflow requires adding a number of parts.

### DIFF
--- a/src/SmartComponents/DriftTable/AddSystem.js
+++ b/src/SmartComponents/DriftTable/AddSystem.js
@@ -19,5 +19,5 @@ export class AddSystem extends Component {
 }
 
 AddSystem.propTypes = {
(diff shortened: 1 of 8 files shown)


src/SmartComponents/DriftTable/modules/operations.js
```diff
@@ -0,0 +1,20 @@
+import axios from 'axios';
```
ehelms: Open question: should we move this file to something like src/api.js ?

johnsonm325: I don't have much knowledge of how to do this or how it all works together, but if we can keep from importing axios in every file by doing this, then I don't see why not.

PR #17, opened 2019-02-05
Show host information aligned to host

ehelms:


beav:
@ehelms it looks like its calling `/r/insights/platform/drift/v0/compare?host_ids=d1d22ce7-4e52-4767-867a-2ccac52485ec`, but it needs to be `host_ids[]`

src/api.js
```diff
@@ -7,6 +7,10 @@ async function get(path, params = {}) {
 }
 
 function getCompare(hostIds = []) {
+    if (!Array.isArray(hostIds)) {
+        hostIds = [ hostIds ];
+    }
```
ehelms: @beav The bug you hit is if a single host_id is passed, the query string parser interprets this as a single value case and converts it to a string rather than an array. This passes it as `host_ids` instead of `host_ids[]` to the backend. This ensures it's always an array and get's converted properly by the params.

PR #18, opened 2019-02-05
Add System Modal

johnsonm325:


src/SmartComponents/modules/reducers.js
```diff
@@ -17,6 +18,11 @@ function compareReducer(state = initialState, action) {
                 ...state,
                 status: action.payload
             };
+        case `${types.OPEN_ADD_SYSTEM_MODAL}_FULFILLED`:
```
ehelms: You don't need the `_FULFILLED` in this case as that originates from the redux-promise-middleware when making async requests. For this, I think you would just want to reference the action directly that will be dispatched.

src/SmartComponents/AddSystemModal/AddSystemModal.js
```diff
@@ -0,0 +1,58 @@
...
+        );
+    }
+}
+
+/*
+function mapStateToProps(state) {
+    return {
+        addSystemModalOpened: state.compareReducer.addSystemModalOpened
+    };
+}
```
ehelms: I think you want this -- that should connect everything.

src/SmartComponents/AddSystemModal/AddSystemModal.js
```diff
@@ -0,0 +1,58 @@
...
+        fetchSystems: (() => dispatch(addSystemModalActions.fetchSystems())),
+    };
+}
+*/
+
+AddSystemModal.propTypes = {
+    showModal: PropTypes.bool,
+};
+
+//export default withRouter (connect(mapStateToProps, mapDispatchToProps)(AddSystemModal));
```
ehelms: Yep! Although you may not need mapDispatchToProps yet.

ehelms:
You'll want to rebase cause of all the little tweaks that were merged yesterday.

src/SmartComponents/AddSystemModal/AddSystemModal.js
```diff
@@ -0,0 +1,42 @@
...
+import { Button, Modal } from '@patternfly/react-core';
+
+export class AddSystemModal extends Component {
+    constructor(props) {
+        super(props);
+    }
+
+    async componentDidMount() {
+        await window.insights.chrome.auth.getUser();
+    }
```
ehelms: This should not be needed since this modal isn't fetching data itself.

PR #21, opened 2019-02-05
Bump patternfly and insights dependencies

ehelms:


ehelms:
An `npm install` suffices. Update does not do what you'd expect.

PR #23, opened 2019-02-05
Highlight different rows

ehelms:


beav:
@ehelms rebase plz

beav:
@ehelms can you make the highlight go all the way across to include the state + name?

ehelms:
The mock ups do not have it extended across the entire row. We can discuss
with Rox.

beav:
if it's the same as the mockup, that is good enough for me. We can discuss during next UX meeting.

PR #27, opened 2019-02-06
Change from host to system nomenclature

ehelms:


beav:
this looks like it will do the trick. https://github.com/RedHatInsights/drift-backend/pull/22 is the matching PR that goes with this.

PR #28, opened 2019-02-06
Add page to show systems table from inventory

ehelms:
Opening this for early stage review and placeholder. This presents the systems as a page with a set of pre-selected systems. This will convert to be embedded in a modal so the component is valid.


ehelms:
Reference for the inventory table integration -- https://github.com/RedHatInsights/insights-frontend-components/blob/master/doc/components/inventory/inventory.md

PR #30, opened 2019-02-07
Add test for Add System Button

johnsonm325:
Let me know if you think this test should be more robust. To be honest, I think we're only testing for proper rendering at this point. Otherwise we could maybe test that clicking the button would call the function, but that seems better tested here: https://github.com/RedHatInsights/drift-frontend/blob/master/src/SmartComponents/DriftPage/DriftTable/DriftTable.js#L101


ehelms:
Good start! I had read about snapshots, good to see them in action.

PR #32, opened 2019-02-08
Set GET params on changing systemIds

ehelms:


beav:
@ehelms tested successfully, thx!

I'll let you hit the merge button, I dont know if you want to do it now or after demo.

PR #37, opened 2019-02-14
Fix broken UI styling by updating insights-frontend

ehelms:


beav:
error from travis is:

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

beav:
@ehelms does it still OOM if you create a PR that just updates the readme? I'm curious if the issue is related to the new version or if it's a travis thing

ryelo:
@beav @ehelms I'm pretty sure this is a Travis error. You should be able to hit travis-ci.org/RedHatInsights/drift-frontend/setts, then in the `Environment Variables` create a new one with the name `NODE_OPTIONS` and a value of `"--max-old-space-size=4096 --max_old_space_size=4096"` (with the quotes is fine). Save, and the kick a new build.

PR #40, opened 2019-02-18
Add drop down for filtering table by SAME, DIFF, and NO DATA states

johnsonm325:


src/SmartComponents/DriftPage/DriftPage.js
```diff
@@ -29,4 +34,10 @@ class DriftPage extends Component {
     }
 }
 
+DriftPage.propTypes = {
+    getFilterByState: PropTypes.func,
+    filter: PropTypes.string,
+    stateFilter: PropTypes.string
+};
```
ehelms: I don't see these props being used anywhere.

johnsonm325: ah yes, you're right. I forgot to remove them while I was trying things out.

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -39,10 +39,17 @@ class DriftTable extends Component {
...
         for (let i = 0; i < data.facts.length; i += 1) {
-            const rowData = this.renderRowData(data.facts[i], data.systems);
-            rows.push(<tr>{ rowData }</tr>);
+            if (this.props.stateFilter === 'none' || this.props.stateFilter === undefined) {
+                rowData = this.renderRowData(data.facts[i], data.systems);
+                rows.push(<tr>{ rowData }</tr>);
+            }
+            else if (this.props.stateFilter === data.facts[i].state) {
+                rowData = this.renderRowData(data.facts[i], data.systems);
+                rows.push(<tr>{ rowData }</tr>);
```
ehelms: This is in both the if and else-if, so you can extract it to the line below the entire block.

src/SmartComponents/DriftPage/FilterDropDown/FilterDropDown.js
```diff
@@ -0,0 +1,87 @@
...
+        this.props.changeFilter(filterData);
+    }
+
+    render() {
+        const dropdownItems = [
+            <DropdownItem key= "none" component="button" onClick={ event =>
+                this.getFilterByState(event, 'none', 'No Filter')
+            }>
+                No Filter
+            </DropdownItem>,
```
ehelms: Can these items be generated programmatically since they share the same structure?

src/SmartComponents/DriftPage/FilterDropDown/FilterDropDown.js
```diff
@@ -0,0 +1,87 @@
...
+FilterDropDown.propTypes = {
+    toggleDropDown: PropTypes.func,
+    toggleFilterDropDown: PropTypes.func,
+    isOpen: PropTypes.bool,
+    filterDropdownOpened: PropTypes.bool,
+    filterByState: PropTypes.func,
+    stateFilter: PropTypes.string,
+    selectedFilter: PropTypes.string,
+    changeFilter: PropTypes.func
+};
```
ehelms: I dont see all of these props being used.

src/SmartComponents/modules/reducers.js
```diff
@@ -54,7 +57,40 @@ function addSystemModalReducer(state = initialState, action) {
...
+    }
+}
+
+function filterByStateReducer(state = initialState, action) {
+    switch (action.type) {
+        case `${types.FILTER_BY_STATE}`:
+            return {
+                ...state,
+                stateFilter: action.payload.filter,
+                selectedFilter: action.payload.selected
```
ehelms: What do these two pieces of data represent? I needing which filter is selected, but what is the other piece of information being passed around?

johnsonm325: stateFilter is the filter that is used to compare against the comparison state text ("none", "SAME", "DIFFERENT", "No Data").

selectedFilter is used to display the selected option from the dropdown menu in the box after a selection has been made and the box collapsed.

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
johnsonm325: Do you mean the
```
rows.push(<tr>{ rowData }</tr>);
```
?

If I do that, I get the following:
image

ehelms: Oh right ,cause there is no default `else` in the if block so you would sometimes just get lots of repetition from the last if-else that triggered.

src/SmartComponents/modules/reducers.js
ehelms: Wouldn't they be one in the same?

johnsonm325: I was handling them differently because one was "DIFFERENT" and the other was "Diff". I can restructure this so we don't have to over-complicate things.

ehelms: We could introduce a mapping so that there is the 'display' property and the 'value' property. Or go with Different for now.

src/SmartComponents/modules/reducers.js
```diff
@@ -3,7 +3,10 @@ import types from './types';
 const initialState = {
     compare: {},
     addSystemModalOpened: false,
-    selectedSystemIds: []
+    selectedSystemIds: [],
+    filterDropdownOpened: false,
+    stateFilter: 'none',
+    selectedFilter: 'No Filter'
```
ehelms: You can drop this now I think?

PR #45, opened 2019-02-28
Add Pagination to table

johnsonm325:
Should work with filtering facts and states. Pagination updates to page 1 when applying new filter.


src/SmartComponents/DriftPage/Pagination/Pagination.js
```diff
@@ -0,0 +1,70 @@
...
+
+    onPerPageSelect(perPage) {
+        const { page } = this.props;
+        const pagination = { page, perPage };
+        this.props.updatePagination(pagination);
+    }
+
+    render() {
+        return (
+            <React.Fragment>
```
ehelms: Is this needed given `<Pagination` is a single element ?

src/SmartComponents/modules/__tests__/reducer-test.js
```diff
@@ -12,23 +12,33 @@ describe('compare reducer', () => {
                 factFilter: '',
                 filterDropdownOpened: false,
                 selectedSystemIds: [],
-                stateFilter: 'all'
+                stateFilter: 'all',
+                filteredData: {},
```
ehelms: For clarity, I think maybe we change `compare` to something like `fullResults` or `fullCompareData` and then this to `filteredCompareData`.

ehelms: Sorry, I know this comment is in a test, but I don't mean it for the test, but generally across the app.

src/SmartComponents/modules/reducers.js
```diff
@@ -6,9 +6,66 @@ const initialState = {
     selectedSystemIds: [],
     filterDropdownOpened: false,
     stateFilter: 'all',
-    factFilter: ''
+    factFilter: '',
+    totalFacts: undefined,
```
ehelms: Initialize this to 0 ?

src/SmartComponents/modules/reducers.js
```diff
@@ -6,9 +6,66 @@ const initialState = {
...
+    page: 1,
+    perPage: 10,
+    filteredData: {}
 };
 
+function getTotal(data, stateFilter, factFilter) {
+    let total = 0;
+
+    for (let i = 0; i < data.facts.length; i++) {
```
ehelms: Note this replicates the logic in `filterCompareData`. Can you call that function directly and take the length?

src/SmartComponents/modules/reducers.js
```diff
@@ -6,9 +6,66 @@ const initialState = {
...
+            }
+        }
+    }
+
+    /* eslint-disable object-shorthand */
+    filteredData = {
+        facts: facts,
+        systems: data.systems
+    };
+    /* eslint-enable object-shorthand */
```
ehelms: Why is this eslint disable needed?

src/SmartComponents/modules/reducers.js
```diff
@@ -22,47 +79,45 @@ function selectedSystems(selectedIds, selectedSystem) {
...
+
     switch (action.type) {
         case `${types.FETCH_COMPARE}_FULFILLED`:
             return {
                 ...state,
                 compare: action.payload,
-                selectedSystemIds: action.payload.systems.map(system => system.id)
+                selectedSystemIds: action.payload.systems.map(system => system.id),
+                filteredData: filterCompareData(action.payload, 1, state.perPage, state.stateFilter, state.factFilter),
+                totalFacts: getTotal(action.payload, state.stateFilter, state.factFilter)
```
ehelms: Any way to save the result of `filterCompareData` to a variable and just take the length?

src/SmartComponents/modules/reducers.js
```diff
@@ -22,47 +79,45 @@ function selectedSystems(selectedIds, selectedSystem) {
 }
 
 function compareReducer(state = initialState, action) {
+    const { payload } = action;
```
ehelms: What's the benefit of saving this off ot a variable vs. calling `action.payload` like we do elsewhere?

src/SmartComponents/modules/reducers.js
```diff
@@ -6,9 +6,66 @@ const initialState = {
...
+                }
+            }
+            else if (stateFilter === data.facts[i].state) {
+                total += 1;
+                if (Math.ceil(total / factsPerPage) === selectedPage) {
+                    facts.push(data.facts[i]);
+                }
+            }
+        }
+    }
```
ehelms: I think this could be cleaner to split into two distinct stages with their own functions:

 1) Filter all the data
 2) Paginate the data

ehelms: This will allow you to unit test cleaner as well. As you can varify pagination logic and filtering logic keeping them more robust over time.

src/SmartComponents/modules/reducers.js
johnsonm325: I must have had it written before in a way that required it to access page and perPage of the payload. I wasn't about to access it via action.payload.page. Now I can!

src/SmartComponents/modules/reducers.js
```diff
@@ -1,14 +1,65 @@
...
+    filteredCompareData: {}
 };
 
+function paginateFilteredData(data, selectedPage, factsPerPage) {
+    let paginatedData = {};
+    let paginatedFacts = [];
+    let divisor = 0;
+
+    for (let i = 0; i < data.facts.length; i++) {
+        divisor++;
```
ehelms: This means that divisor starts at `1` and is incremented on every iteration right? The `i` variable starts at `0` and increments every iteration. Feels like you could simplify it down to one variable?

src/SmartComponents/modules/reducers.js
```diff
@@ -1,14 +1,65 @@
...
+
+    paginatedData = {
+        facts: paginatedFacts,
+        systems: data.systems
+    };
+
+    return paginatedData;
+}
+
+function filterCompareData(data, stateFilter, factFilter) {
```
ehelms: If we just pass in `facts` here then we simplify this function and its scope of knowledge. That is, it only has to care about a facts data structure not a hash that has a `facts` key.

src/SmartComponents/modules/reducers.js
johnsonm325: yes, but I know it's not standard to modify the for loop counter and it is standard to start at 0. But, if you're good with it, I'll just do math on i+1.

src/SmartComponents/modules/reducers.js
johnsonm325: Correct me if I'm wrong, but since we populate the DriftTable columns using the data.systems (https://github.com/RedHatInsights/drift-frontend/blob/master/src/SmartComponents/DriftPage/DriftTable/DriftTable.js#L51), then we'd want to build that filteredData with the facts and systems. If we only pass in facts, then we have to find a more convoluted way to build the filtered data outside of the function.

src/SmartComponents/modules/reducers.js
ehelms: Sounds good to me

src/SmartComponents/modules/reducers.js
ehelms: Correct, but this function is *only* concerned with taking and filtering the set of facts into a subset of facts. Note that it only operates on `data.facts` so the `data.` part can be factored out.

src/SmartComponents/modules/reducers.js
```diff
@@ -1,14 +1,51 @@
...
+            paginatedFacts.push(data[i]);
+        }
+    }
+
+    return paginatedFacts;
+}
+
+function filterCompareData(data, stateFilter, factFilter) {
+    if (data === undefined) {
+        return {};
```
ehelms: Note this is a hash, but normally this function returns an array. I think we could drop this if statement all together.

ehelms:
Needs a rebase and one small comment.

PR #46, opened 2019-02-28
Update host styling to add background and last updated

ehelms:


ehelms:
@johnsonm325 Looks like I need to generate a new snapshot... how do I do that?

PR #47, opened 2019-03-01
Add empty state

johnsonm325:
Chose to add the empty state inside of the table. Mostly because it's the easiest way to do it, but can figure out how to populate the entire page with it if that won't work, we'd just need to do some restructuring.

### DIFF
--- a/src/SmartComponents/DriftPage/DriftPage.js
+++ b/src/SmartComponents/DriftPage/DriftPage.js
@@ -5,7 +5,6 @@ import { Card, CardBody, Grid, GridItem } from '@patternfly/react-core';
 import DriftTable from './DriftTable/DriftTable';
 import FilterDropDown from './FilterDropDown/FilterDropDown';
 import SearchBar from './SearchBar/SearchBar';
-import TablePagination from './Pagination/Pagination';
 
 class DriftPage extends Component {
     constructor(props) {
@@ -32,7 +31,6 @@ class DriftPage extends Component {
                             <div>
                                 <DriftTable />
                             </div>
-                            <TablePagination />
                         </CardBody>
                     </Card>
                 </Main>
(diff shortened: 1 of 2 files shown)


src/SmartComponents/DriftPage/DriftPage.js
```diff
@@ -1,42 +1,89 @@
...
+                                        { /*<RenderPage { ...this.props.compare } />*/ }
+                                        <EmptyTableState />
+                                    </center>
+                                </div>
+                            </CardBody>
+                        </Card>
+                    </Main>
+                </React.Fragment>
+            );
+        }
```
ehelms: There's a lot of overlap here. Can we reduce the if/else to just the parts that are changing?

johnsonm325: I'm struggling with the best way to handle this then. You can't put if/else statements inside of JSX (inside of the return). Conditional rendering is handled like so: https://reactjs.org/docs/conditional-rendering.html but we're looking at rending 1 component vs. 3 components, so that complicates things.

I'm open to thoughts on this.

ehelms: Only other thought I have is to introduce two new components:

 * DriftEmptyStatePage
 * DriftTablePage

That could be overkill for now.

ehelms:
Gonna need a rebase

src/SmartComponents/DriftPage/DriftPage.js
```diff
@@ -15,7 +15,7 @@ class DriftPage extends Component {
     render() {
         return (
             <React.Fragment>
-                <PageHeader>
+                *<PageHeader>
```
ehelms: Is the `*` intentional?

johnsonm325: nope, was a mistake! I was wondering where that came from...

ehelms:
Pagination is still showing due to it living in the DriftPage I think. We may need to discuss this further as I think this and my loading design diverge.

PR #49, opened 2019-03-07
Add loading table skeleton

ehelms:


ehelms:
This is updated to be rebased on the empty state. I've taken the original design and re-worked it to match other render functionality we had already and how I'd approached the loading. Let me know what you think about this overall @johnsonm325

johnsonm325:
Do we want to include the "Add System" button in the loading screen?

image

ehelms:
Hrmm, fair thought. I will see what it takes to pull that off.

ehelms:
Updated to fix a bug and set the button to loading as well.

PR #51, opened 2019-03-08
Re-factor to an AddSystemButton

ehelms:
The button for adding systems is used in two places and should
reflect the same functionality and styling. This moves the styling
for the table into the table and re-uses the same button across empty
state and active table.


ehelms:
I should add a test as well but I wanted to get your thoughts @johnsonm325

PR #54, opened 2019-03-13
Remove host

ehelms:


ehelms:
This work requires and is based upon the re-factoring work from https://github.com/RedHatInsights/drift-frontend/pull/53

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -17,30 +17,46 @@ import AddSystemButton from '../AddSystemButton/AddSystemButton';
...
+        this.systemIds = this.systemIds.filter(item => item !== undefined);
     }
 
     formatDate(dateString) {
         let date = new Date(dateString);
         return date.toLocaleDateString() + ' ' + date.toLocaleTimeString();
     }
 
+    removeSystem(systemId) {
+        let ids = this.systemIds.filter(item => item !== systemId);
```
johnsonm325: If you:
- load the blank state
- choose 2+ systems
- immediately click one of the Xs

You lose all of the systems you selected. systemIds at this point is empty, so it runs a compare on nothing. Refreshing the page with the system Ids in the url will load them into systemIds.

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -17,30 +17,46 @@ import AddSystemButton from '../AddSystemButton/AddSystemButton';
...
+        if (ids.length > 0) {
+            this.fetchCompare(ids);
```
johnsonm325: Here, we'll need to remove the systemId from the systemIds array. If you:

- Choose 2 systems
- Click one of the Xs (removes as expected)
- Click the X on the other card

You get the other card back that you originally clicked the X on. It's because It's still in the systemIds array.

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
ehelms: I'm having a hard time re-producing this issue. The loading screen covers up the 'x' to click it for me preventing any issues?

ehelms:
Updated addressing the issues identified. I also fixed a bug where systems with no facts (rare but possible) trigger the empty state to happen.

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
johnsonm325: I think this was cleared up with your fix.

PR #56, opened 2019-03-14
Add alphabetical sorting of fact name

johnsonm325:
Sort starts as "none". Clicking the sort button will automatically set it to ascending. Clicking again will set to descending and will flip back and forth between ascending and descending. Does not reset to "none" unless you refresh the page.

I know this will change a bit with #53.


src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -73,25 +73,25 @@ class DriftTable extends Component {
     }
 
     renderRowData(fact, systems) {
-        let td = [];
+        let tr = [];
 
-        td.push(<td>{ fact.name }</td>);
-        td.push(<td className="fact-state"><StateIcon factState={ fact.state }/></td>);
+        tr.push(<th>{ fact.name }</th>);
+        tr.push(<th className="fact-state"><StateIcon factState={ fact.state }/></th>);
```
ehelms: Why change these `table data` cells to `table header` cells?

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -73,25 +73,25 @@ class DriftTable extends Component {
     }
 
     renderRowData(fact, systems) {
-        let td = [];
+        let tr = [];
```
ehelms: Maybe we call this `row` ?

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -118,6 +118,18 @@ class DriftTable extends Component {
         return row;
     }
 
+    renderSortButton(props) {
+        if (props.sort === 'none') {
```
ehelms: The `props` aspect here is an unnecessary abstraction since you never access any other key of the props hash. If you pass in just `sort` then the function is cleaner.

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -118,6 +118,18 @@ class DriftTable extends Component {
...
+            return <SortIcon onClick={ () => this.props.toggleFactSort('asc') }/>;
+        }
+        else if (props.sort === 'asc') {
+            return <AngleUpIcon onClick={ () => this.props.toggleFactSort('desc') }/>;
+        }
+        else if (props.sort === 'desc') {
+            return <AngleDownIcon onClick={ () => this.props.toggleFactSort('asc') }/>;
+        }
```
ehelms: I think it can be a bit cleaner to set these equal to a variable and do a single return at the end.

src/SmartComponents/modules/reducers.js
```diff
@@ -40,6 +41,17 @@ function filterCompareData(data, stateFilter, factFilter) {
...
+    if (sort === 'asc') {
+        filteredFacts.sort(function(a, b) {
+            return a.name > b.name;
+        });
+    }
+    else if (sort === 'desc') {
+        filteredFacts.sort(function(a, b) {
+            return b.name > a.name;
+        });
+    }
```
ehelms: I think this overloads the filtering function. Any data should be sortable, so splitting it out to a function just like filtering and pagination makes it more atomic and easier to test.

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
johnsonm325: It must have carried over from my horizontal scroll PR. I must have made a mistake somewhere when creating this branch. I'll revert it.

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -116,6 +116,22 @@ class DriftTable extends Component {
...
+        let sortIcon;
+
+        if (sort === 'none') {
+            sortIcon = <SortIcon onClick={ () => this.props.toggleFactSort('asc') }/>;
+        }
+        else if (sort === 'asc') {
+            sortIcon = <AngleUpIcon onClick={ () => this.props.toggleFactSort('desc') }/>;
+        }
+        else if (sort === 'desc') {
+            sortIcon = <AngleDownIcon onClick={ () => this.props.toggleFactSort('asc') }/>;
```
ehelms: Should this match `'desc` ?

src/SmartComponents/modules/reducers.js
```diff
@@ -11,11 +11,12 @@ const initialState = {
     factFilter: '',
     totalFacts: 0,
     page: 1,
+    sort: 'none',
     perPage: 10,
     loading: false
 };
 
-function paginateFilteredData(data, selectedPage, factsPerPage) {
+function paginateSortedData(data, selectedPage, factsPerPage) {
```
ehelms: This function doesn't care if the data is sorted or filtered right? so this could just be `paginateData` ?

src/SmartComponents/modules/reducers.js
```diff
@@ -44,6 +45,21 @@ function filterCompareData(data, stateFilter, factFilter) {
     return filteredFacts;
 }
 
+function sortFilteredData(filteredFacts, sort) {
```
ehelms: Similar comment, this function only cares that it is handed data.

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
johnsonm325: Not sure what you mean by this...

src/SmartComponents/modules/__tests__/reducer-test.js
```diff
@@ -12,6 +12,7 @@ describe('compare reducer', () => {
                 factFilter: '',
                 filterDropdownOpened: false,
                 selectedSystemIds: [],
+                sort: 'none',
```
ehelms: asc

src/SmartComponents/modules/__tests__/reducer-test.js
```diff
@@ -106,6 +107,7 @@ describe('add system modal reducer', () => {
...
```
ehelms: asc

ehelms:
Looks good once test failures are fixed

PR #58, opened 2019-03-18
Sort selected systems at the top

ehelms:


johnsonm325:
If we could get a rebase on this one so we can test on the new cloud env, that would be great!

ehelms:
@johnsonm325 Rebased, happy testing

PR #61, opened 2019-03-22
Fix CI environment path name

johnsonm325:


beav:
@johnsonm325 I added this originally from https://github.com/RedHatInsights/insights-frontend-starter-app/pull/63/files. I'm OK with removing it to see if it's not needed or causing issues.

PR #65, opened 2019-03-26
Add custom_release to push to ci-stable on master commits

ehelms:


ehelms:
Makes use of this ability in deployment scripts -- https://github.com/RedHatInsights/insights-frontend-builder-common/blob/master/src/after_success.sh#L23

beav:
@ehelms change looks good to me, but how did this work previously if we weren't specifying `ci-beta`?

ehelms:
@beav this logic was triggered https://github.com/RedHatInsights/insights-frontend-builder-common/blob/master/src/after_success.sh#L28

PR #66, opened 2019-03-27
Add Export Button

johnsonm325:
Clicking the export button will automatically download a CSV file with no regard to pagination. Other filters will remain in effect.


johnsonm325:
has failed tests...

src/SmartComponents/modules/reducers.js
```diff
@@ -88,6 +89,73 @@ function selectedSystems(selectedIds, selectedSystem) {
     return selectedIds;
 }
 
+function convertFactsToCSV(data, systems) {
+    let result;
+    let ctr;
```
ehelms: What does `ctr` stand for?

src/SmartComponents/modules/reducers.js
```diff
@@ -144,6 +214,7 @@ function compareReducer(state = initialState, action) {
                 stateFilter: action.payload,
                 page: 1,
                 filteredCompareData: paginatedFacts,
+                dataToConvert: sortedFacts,
```
ehelms: Should we store this as `sortedFilteredFacts` ? I find `dataToConvert` not as obvious as to what it means.

src/SmartComponents/modules/reducers.js
```diff
@@ -88,6 +89,73 @@ function selectedSystems(selectedIds, selectedSystem) {
...
+    columnDelimiter = data.columnDelimiter || ',';
+    lineDelimiter = data.lineDelimiter || '\n';
+
+    result = 'Fact,State,';
+    result += systems.join(columnDelimiter);
+    result += lineDelimiter;
+
+    keys = Object.keys(data[0]);
+
+    data.forEach(function(fact) {
```
ehelms: I think you can do:

```
data.forEach(function(fact, index) {
```

And avoid the need for `ctr` object

src/SmartComponents/modules/reducers.js
```diff
@@ -88,6 +89,70 @@ function selectedSystems(selectedIds, selectedSystem) {
...
+    let data;
+    let filename;
+    let link;
+    let csv = convertFactsToCSV(driftData, systems);
+
+    if (csv === null) {
+        return;
+    }
+
+    filename = 'export.csv';
```
ehelms: Maybe name it `system-comparison-export-%date.csv` so they can export them easily without clashes and know what the export is for?

src/SmartComponents/modules/reducers.js
```diff
@@ -88,6 +89,70 @@ function selectedSystems(selectedIds, selectedSystem) {
...
+    if (data === null || !data.length) {
+        return null;
+    }
+
+    columnDelimiter = data.columnDelimiter || ',';
+    lineDelimiter = data.lineDelimiter || '\n';
+
+    result = 'Fact,State,';
+    result += systems.join(columnDelimiter);
+    result += lineDelimiter;
```
ehelms: This builds up the 'headers', is that correct? If so, it would help my readability to have this as a separate variable and then combine them at the end, e.g.  headers + result

ehelms:
Maybe we disable or hide the export button in the empty state?

ehelms:
The export sets the system based on its UUID currently, which I think would be hard for a user to follow which system is which. I think we should change this to be the system name instead.

src/SmartComponents/DriftPage/ExportButton/ExportButton.js
```diff
@@ -0,0 +1,34 @@
...
+import { Button } from '@patternfly/react-core';
+
+class ExportButton extends Component {
+    constructor(props) {
+        super(props);
+    }
+
+    render() {
+        return (
+            <Button
```
ehelms: Can we set this to `float: right` so it is on the right side of the grid element?

PR #77, opened 2019-04-05
Add categorization

johnsonm325:
Click arrow to expand and show child facts. Fact filtering filters on facts only, not fact headers (parents).


src/SmartComponents/modules/reducers.js
```diff
@@ -153,10 +186,23 @@ function downloadCSV(driftData, systems) {
     link.dispatchEvent(new MouseEvent(`click`, { bubbles: true, cancelable: true, view: window }));
 }
 
+function toggleExpandedRow(expandedRows, factName) {
+    if (expandedRows.includes(factName)) {
+        expandedRows = _.remove(expandedRows, function(fact) {
```
ehelms: I think you can reverse this logic by using built in `filter` function and avoid needing lodash.

ehelms:
Thoughts from testing:

 * Can you make it so that the cursor changes to a hand when hovering over the carat?
 * The expandable rows do not extend all the way to the right and are missing bottom borders (See pic below)
 * I think we could save some space and indent the expanded rows by half of what they are today
 * If I type in 'systemd' to search, I still see 'cpu flags' even though there is no data underneath it. I think we should hide top level categories if there is no data underneath

Screenshot from 2019-04-05 09-51-44

src/App.scss
```diff
@@ -102,6 +102,9 @@
     min-width: 60px;
     max-width: 60px;
   }
+  .child-row {
+    margin-left: 5em;
```
ehelms: Maybe drop this to 2 or 3 em?

johnsonm325:
Last question on this PR: If we apply a fact filter to the data and a category that was previously set to INCOMPLETE is suddenly filled with all SAME data, do we want to update the category state to be SAME, or leave it as INCOMPLETE?

johnsonm325:
This breaks with CSV update. I need to fix that before we merge

beav:
this depends on https://github.com/RedHatInsights/drift-backend/pull/68, both need to merge around the same time when merging.

PR #78, opened 2019-04-05
Fix possible null values when exporting CSV

ehelms:


beav:
@ehelms when is the value null? the PR looks good to me but i want to make sure it's not masking a backend bug

johnsonm325:
> @ehelms when is the value null? the PR looks good to me but i want to make sure it's not masking a backend bug

We get null back when a system's value is INCOMPLETE_DATA, which makes sense. We just have to catch it on the frontend.

PR #91, opened 2019-04-18
Add bottom blue border to fact column header

johnsonm325:


src/App.scss
```diff
@@ -30,9 +30,13 @@
     padding-right: 2px;
   }
 
+  .fact-header-cell {
+    td {
+      border-bottom: 3px solid #0000FF;
```
ehelms: We want to use `#39a5dc` for this color. Can you add coloring of the caret as well?

PR #108, opened 2019-04-30
Enable select all functionality to work on Systems modal

ehelms:
This necesitates a re-work of how select functionality is handled
throughout the application and simplifies overall design.


beav:
@ehelms this works, but if you have more than 48 UUIDs you'll hit a URL length limit.

If I added POST support to the backend for sending UUIDs, could you switch the frontend over?

PR #112, opened 2019-04-30
Check url for systems before rendering

johnsonm325:


src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -306,7 +306,7 @@ class DriftTable extends Component {
                     showModal={ this.props.addSystemModalOpened }
                     confirmModal={ this.fetchCompare }
                 />
-                { systems.length > 0 || loading ?
+                { systems.length > 0 || loading || location.search !== '' ?
```
ehelms: I think it could be cleaner to check if systemIds is set here?

```
this.systemIds.length !== 0
```

beav:
I wasn't able to repro the bug without the patch. Do I just go to the drift app with a system_id in place on the url?

ehelms:
@beav yep

PR #114, opened 2019-04-30
Switch to post UUIDs to backend

ehelms:


johnsonm325:
I'm assuming that the backend will be changing to accept this post? If so, it looks good to me. After I get the confirmation on this, I can approve.

beav:
this change should not be merged for 2 weeks. I would recommend closing and re-opening then.

beav:
I used this to test the backend changes and it worked for me.

PR #118, opened 2019-05-01
Add pagination to bottom of table also

ehelms:


beav:
I still only see the pagination at the top with this patch. Do I need to do anything special after applying?

ehelms:
@beav Can you paste a screenshot ?

johnsonm325:
@ehelms I can see it on the top and the bottom. But the new pagination at the bottom is visible on empty state and loading.

beav:
> @beav Can you paste a screenshot ?

It's there now for me :man_shrugging: I guess something was cached

PR #120, opened 2019-05-01
Closing error alert properly restores previous state

johnsonm325:
Removed resetSelectedIds from ErrorAlert component as it was removed from the code base in a previous PR: https://github.com/RedHatInsights/drift-frontend/pull/108


beav:
when I get a 400 and close it out, it sends me back to the empty state. is that the expected behavior?

johnsonm325:
@beav Yes, that is expected behavior. It is basically keeping what you previously compared in the table and is restoring that. I'm assuming before you ran the compare, you had an empty table.

If you run a compare, mess up the backend and run it again, and then try to hit the api (500), you'll get the error and on close you should be kicked back to the original compare you ran.

PR #124, opened 2019-05-03
On running fetch, the submitted ids comes from the selectedSystemIds

johnsonm325:
The bug is, when you select a system from the modal on page 1, then go to page 2 and select another system and click compare, you only submit the selected system of the page you are currently on. By changing to selectedSystemIds, we are submitting all systems that have been selected in the modal. Systems are added and taken away from the selectedSystemIds array each time the modal is opened.


beav:
was able to repro bug, testing w/ patch now

beav:
works for me

PR #126, opened 2019-05-03
Actively sorted column includes blue bottom border

johnsonm325:
Bug in code was causing it to not render, but now it is.


johnsonm325:
Looks like this was already working in ci environment. But the prod env does not show the blue bar. May want to investigate that first.

PR #131, opened 2019-05-07
Add state sort to table

johnsonm325:
Fact column is always sorted by either ascending or descending.
State column is turned off by default; and then can be flipped from:
off -> ascending
ascending -> descending
descending -> off

The column is rendered with a blue bar underneath the header title based on which sort button was last clicked. If you flip between sorts, the blue bar will render differently with each click.

State sort takes priority when turned on. Click the fact sort while the state sort is turned on will sort the fact name alphabetically within the state sort. To get a pure alphabetical sort, turn the state filter off completely. Here's the breakdown of state sorting:
Ascending = same -> incomplete data -> different
Descending = different -> incomplete data -> same


src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -215,19 +215,78 @@ class DriftTable extends Component {
...
 
         return sortIcon;
     }
 
+    toggleSort(sortType, sort) {
+        if (sort === 'asc') {
+            if (sortType === 'fact') {
+                this.props.toggleFactSort('desc');
+            } else {
+                this.props.toggleStateSort('desc');
```
ehelms: If the `sort` that is passed in is `asc` why do we pass `desc` into the function? My intuition says it should be just `sort` that is passed in. What am I missing?

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -215,19 +215,78 @@ class DriftTable extends Component {
...
+        }
+
+        this.props.toggleActiveSort(sortType);
+    }
+
+    renderHeaderRow(systems, loading) {
+        if (this.props.activeSort === 'fact' || this.props.stateSort === '') {
+            return (
+                <tr className="sticky-column-header">
+                    <th className="active-sort fact-header sticky-column fixed-column-1">
```
ehelms: As far as I can tell, the only difference between the if-else is the addition of this class on one of the two table headers. Can we drop the if-else and make inclusion of this class the part triggered by logic?

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
johnsonm325: So, if the current sort is ascending, and someone clicks it to change it to descending, I'm passing 'desc' to the action so we can update the sort in the reducer by simply doing this: https://github.com/RedHatInsights/drift-frontend/pull/131/files#diff-d585a54c07f675fb1ae7d1906b05671eR391

I can change it to put the logic you mention above into the reducer instead, if you think that makes more sense.

src/SmartComponents/modules/actions.js
```diff
@@ -47,9 +47,35 @@ function toggleFilterDropDown() {
 }
 
 function toggleFactSort(sortType) {
+    let newSort;
+
+    if (sortType === 'asc') {
```
ehelms: Not required, but we might want to consider moving these to constants that can be referenced.

src/SmartComponents/modules/actions.js
```diff
@@ -47,9 +47,35 @@ function toggleFilterDropDown() {
...
+        newSort = 'desc';
+    } else {
+        newSort = 'asc';
+    }
+
     return {
         type: types.TOGGLE_FACT_SORT,
-        payload: sortType
+        payload: newSort
```
ehelms: You could consider doing

```
payload: sortType === 'asc' ? 'desc' : 'asc'
```

src/SmartComponents/modules/actions.js
ehelms: I think it would help me with readability if `sortType` was something like `previousSort` or `currentSort` to indicate the value that is coming in is the old value that we need to translate to the new.

PR #133, opened 2019-05-08
Add ability to select multiple state filters

johnsonm325:
By default, all state filters are selected.


johnsonm325:
Gotta fix tests. Forgot about that

src/SmartComponents/DriftPage/FilterDropDown/FilterDropDown.js
```diff
@@ -9,44 +9,43 @@ class FilterDropDown extends Component {
...
-            this.createDropdownItem('Different', 'DIFFERENT'),
-            this.createDropdownItem('Incomplete Data', 'INCOMPLETE_DATA')
+            this.createDropdownItem('same-check', 'Same', 'SAME', same),
+            this.createDropdownItem('different-check', 'Different', 'DIFFERENT', different),
+            this.createDropdownItem('incomplete-check', 'Incomplete Data', 'INCOMPLETE_DATA', incomplete)
         ];
 
+        let selectedViews = '';
+        for (let i = 0; i < stateFilters.length; i++) {
+            selectedViews += (stateFilters[i].charAt(0).toUpperCase() + stateFilters[i].slice(1).toLowerCase()).replace(/_/g, ' ');
```
ehelms: Given this and the above, I feel like this might be the time to consider a re-factor to a data structure that holds the different filters. Possibly defined in the state?

src/SmartComponents/DriftPage/FilterDropDown/FilterDropDown.js
```diff
@@ -9,44 +9,45 @@ class FilterDropDown extends Component {
...
+
+        let selectedViews = '';
+        for (let i = 0; i < stateFilters.length; i++) {
+            if (stateFilters[i].checked) {
+                selectedViews += stateFilters[i].display;
+                if ((i + 1) < stateFilters.length) {
+                    selectedViews += ', ';
+                }
+            }
+        }
```
ehelms: All this logic here and above feels like they should exist in functions rather than render

src/SmartComponents/DriftPage/FilterDropDown/FilterDropDown.js
```diff
@@ -9,44 +9,45 @@ class FilterDropDown extends Component {
...
-            this.createDropdownItem('Same', 'SAME'),
-            this.createDropdownItem('Different', 'DIFFERENT'),
-            this.createDropdownItem('Incomplete Data', 'INCOMPLETE_DATA')
-        ];
+        const { stateFilters } = this.props;
+        const dropdownItems = [];
+        stateFilters.forEach(function(stateFilter) {
+            let dropdownItem = this.createDropdownItem(stateFilter);
+            dropdownItems.push(dropdownItem);
+        }.bind(this));
```
ehelms: Why does this have to bind?

src/SmartComponents/modules/reducers.js
```diff
@@ -8,7 +8,11 @@ const initialState = {
     previousStateSystems: [],
     addSystemModalOpened: false,
     filterDropdownOpened: false,
-    stateFilter: 'all',
+    stateFilters: [
+        { filter: 'SAME', display: 'Same', checked: true },
```
ehelms: I tend to see `selected` as the word used in these situations.

src/SmartComponents/modules/reducers.js
```diff
@@ -81,15 +89,18 @@ function filterCompareData(data, stateFilter, factFilter, newExpandedRows) {
...
-function filterComparisons(comparisons, stateFilter, factFilter) {
+function filterComparisons(comparisons, stateFilters, factFilter) {
     let filteredComparisons = [];
+    let isStateChecked;
 
     for (let i = 0; i < comparisons.length; i++) {
-        if (comparisons[i].name.includes(factFilter)) {
-            if (stateFilter.toLowerCase() === 'all' || stateFilter === undefined) {
-                filteredComparisons.push(comparisons[i]);
+        isStateChecked = stateFilters.find(function(stateFilter) {
```
ehelms: I've seen this logic in two places, might be a candidate for a helper function.

src/SmartComponents/modules/reducers.js
```diff
@@ -234,13 +245,28 @@ function toggleExpandedRow(expandedRows, factName) {
...
+function updateStateFilters(stateFilters, updatedStateFilter) {
+    let newStateFilters = [];
+
+    stateFilters.forEach(function (stateFilter) {
+        if (stateFilter.filter === updatedStateFilter.filter) {
+            newStateFilters.push(updatedStateFilter);
+        } else {
+            newStateFilters.push(stateFilter);
+        }
+    });
```
ehelms: I'm surprised by this logic. Based on your data structure, I was assuming that updating the set of filters would be setting 'selected' to true or false based on whether a user selected or unselected a filter.

src/SmartComponents/DriftPage/FilterDropDown/FilterDropDown.js
johnsonm325: It doesn't know what this.createDropdownItem is if you don't. It says "this is undefined"

src/SmartComponents/modules/reducers.js
johnsonm325: So, this is annoying. If I change the function like so:
```
    stateFilters.forEach(function (stateFilter) {
        if (stateFilter.filter === updatedStateFilter.filter) {
            stateFilter.selected = updatedStateFilter.selected;
        }
    });

    return stateFilters;
```
the checkboxes don't check and uncheck until I close the dropdown box.

src/SmartComponents/DriftPage/FilterDropDown/FilterDropDown.js
ehelms: I thought it should work if you drop the `this` all together.

johnsonm325: If I do that, it says "createDropdownItem" is undefined.

ehelms:
Looks good. The checkboxes are tight against the left hand side of the dropdown. Could you add some padding ?

src/SmartComponents/modules/reducers.js
```diff
@@ -37,18 +41,33 @@ function paginateData(data, selectedPage, factsPerPage) {
...
+
+    return isStateSelected;
+}
+
+function filterCompareData(data, stateFilters, factFilter, newExpandedRows) {
     let filteredFacts = [];
     let filteredComparisons = [];
+    let isStateChecked;
```
ehelms: Should we stick to `isStateSelected` lingo here?

src/SmartComponents/modules/reducers.js
```diff
@@ -37,18 +41,33 @@ function paginateData(data, selectedPage, factsPerPage) {
     return paginatedFacts;
 }
 
-function filterCompareData(data, stateFilter, factFilter, newExpandedRows) {
+function findIsStateChecked(state, stateFilters) {
```
ehelms: `isStateSelected` ?

PR #137, opened 2019-05-24
Add action to remove all systems

johnsonm325:


src/Utilities/SetHistory.js
```diff
@@ -0,0 +1,18 @@
+//import { withRouter } from 'react-router-dom';
```
ehelms: Can drop this I assume?

src/Utilities/SetHistory.js
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+    /*eslint-disable camelcase*/
+    history.push({
+        search: '?' + queryString.stringify({ system_ids: systemIds })
+    });
+    /*eslint-enable camelcase*/
+}
+
+//export default withRouter(setHistory);
```
ehelms: Can drop this I assume?

johnsonm325: OMG!

src/Utilities/SetHistory.js
```diff
@@ -0,0 +1,17 @@
+import queryString from 'query-string';
+
+export function setHistory(history, systemIds, clearState) {
+    if (clearState === true) {
```
ehelms: I am not seeing the need for clearState. This function is not additive so it will always reset the set of systemIds on the history. I would consider performing this based on an empty set of systemIds or create a clearHistory function. This might be useful to consider in light of wanting to also store filters on the URL.

src/SmartComponents/ErrorAlert/ErrorAlert.js
```diff
@@ -3,33 +3,20 @@ import PropTypes from 'prop-types';
...
-        this.props.history.push({
-            search: ''
-        });
-
-        /*eslint-disable camelcase*/
-        this.props.history.push({
-            search: '?' + queryString.stringify({ system_ids: this.props.previousStateSystems.map(system => system.id) })
-        });
-        /*eslint-enable camelcase*/
+        setHistory(this.props.history, this.props.previousStateSystems.map(system => system.id), true);
```
ehelms: See comment about this function not being additive.

src/SmartComponents/DriftPage/ActionKebab/ActionKebab.js
```diff
@@ -0,0 +1,69 @@
...
+
+    exportSelect() {
+        this.props.toggleKebab();
+        this.props.exportToCSV();
+    }
+
+    removeSystemsSelect() {
+        this.props.toggleKebab();
+        this.props.removeSystems();
+        this.props.history.push({
```
ehelms: Good use of a clearHistory function

ehelms:
Looks like this needs a rebase

PR #138, opened 2019-05-28
Use frontend components monorepo and update patternfly deps

karelhala:
We have recently switched to monorepo under https://github.com/RedHatInsights/frontend-components. This PR updates deps to use this new packages and updates PF deps.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3246,27 +3246,6 @@
         }
       }
     },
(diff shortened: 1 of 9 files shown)


ehelms:
https://github.com/RedHatInsights/drift-frontend/pull/141 includes and replaces this to fix the failing tests

PR #140, opened 2019-05-28
Refactor addSystemModalReducer

johnsonm325:
Moved addSystemModalReducer into it's own file. Will do this with other reducers if we are ok with this setup.

### DIFF
--- a/src/SmartComponents/modules/addSystemModalReducer.js
+++ b/src/SmartComponents/modules/addSystemModalReducer.js
@@ -1,0 +1,24 @@
+import types from './types';
+
+const initialState = {
+    addSystemModalOpened: false
+};
+
+function addSystemModalReducer(state = initialState, action) {
+    switch (action.type) {
+        case `${types.OPEN_ADD_SYSTEM_MODAL}`:
+            return {
+                ...state,
+                addSystemModalOpened: !state.addSystemModalOpened
+            };
+
+        default:
+            return {
+                ...state
+            };
+    }
+}
+
+export default {
+    addSystemModalReducer
+};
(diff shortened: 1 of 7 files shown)


src/SmartComponents/modules/addSystemModalReducer.js
```diff
@@ -0,0 +1,24 @@
+import types from './types';
```
ehelms: Given this reducer is specific now to addSystemModal I feel it should live with the addSystemModal code. This would isolate it and create more of a full fledged 'component'. And by this I mean this file, not combining all code into a single file. The original idea I was using for layouts is called *duck* [1]. I'd love to get your thoughts on this abstraction for laying out our code. I'm open to other ideas as well.

[1] https://www.freecodecamp.org/news/scaling-your-redux-app-with-ducks-6115955638be/

johnsonm325: So, I'm confused. You want to go with the ducks approach? If so, they go for a function-first approach, which means splitting things out by action/reducers/types/etc. But it seems like you're asking me to go with a feature-first (or component-first) approach.

Perhaps I'm misunderstanding what you're trying to say here. :)

ehelms: My reply got lost somewhere. I read up on a ducks at scale approach which argues for the pattern just applied across components. If for example, a reducer only applies to a single component, you isolate its code to that components in the same structure. I could live with a top level ducks approach to start with and only break it up as we need.

johnsonm325: In that case, this reducer would live with the component, right? Are we good to go with this PR then?

PR #143, opened 2019-05-31
Use auto width for system dialog

karelhala:
### Before
Screenshot from 2019-05-31 16-33-51

### After
Screenshot from 2019-05-31 16-31-44


karelhala:
ping @ehelms

PR #146, opened 2019-06-06
Refactor reducers

johnsonm325:


src/SmartComponents/AddSystemModal/AddSystemModal.js
```diff
@@ -71,7 +71,7 @@ AddSystemModal.propTypes = {
 
 function mapStateToProps(state) {
     return {
-        systems: state.compareReducer.systems,
+        systems: state.appState.compareState.systems,
```
ehelms: download

src/SmartComponents/AddSystemModal/addSystemModalReducer.js
```diff
@@ -0,0 +1,15 @@
+import types from '../modules/types';
```
ehelms: I would expect this type to live here given its related to this reducer?

src/SmartComponents/DriftPage/ActionKebab/actionKebabReducer.js
```diff
@@ -0,0 +1,14 @@
+import types from '../../modules/types';
```
ehelms: Similar comment

src/SmartComponents/DriftPage/AddSystemButton/addSystemModalReducer.js
```diff
@@ -0,0 +1,14 @@
+import types from '../../modules/types';
```
ehelms: Similar

src/store/index.js
```diff
@@ -16,11 +16,7 @@ export function init (...middleware) {
     ]);
 
     registry.register({
-        compareReducer: reducers.compareReducer,
-        addSystemModalReducer: reducers.addSystemModalReducer,
-        errorAlertReducer: reducers.errorAlertReducer,
-        filterDropdownReducer: reducers.filterDropdownReducer,
-        actionKebabReducer: reducers.actionKebabReducer
+        appState: appReducer
```
ehelms: Do I understand this correctly that all of the reducers now live within a single slice of the state known as `appState` ?

src/SmartComponents/modules/appReducer.js
```diff
@@ -0,0 +1,14 @@
+import { combineReducers } from 'redux';
+import { compareReducer } from './reducers';
+import { addSystemModalReducer } from '../DriftPage/AddSystemButton/addSystemModalReducer';
+import { errorAlertReducer } from '../ErrorAlert/errorAlertReducer';
+import { filterDropdownReducer } from '../DriftPage/FilterDropDown/filterDropdownReducer';
+import { actionKebabReducer } from '../DriftPage/ActionKebab/actionKebabReducer';
+
+export const appReducer = combineReducers({
+    compareState: compareReducer,
```
ehelms: Why does the name of the reducer change here from `compareReducer` to `compareState` ?

johnsonm325: So, I think this will answer this question and the one above:

image

Yes, all slices live within a single state known as appState (GO MOUNTAINNEERS!).
I renamed compareReducer to compareState to better represent what the object was. The compareReducer is the function above, while the object compareReducer returns is the compareState.

I would ultimately like to get rid of the compareState and appState altogether so everything lives within state.<whatever>.

src/SmartComponents/AddSystemModal/AddSystemModal.js
johnsonm325: APP!!!

STATE!!!

APP!!!

STATE!!!

johnsonm325:
So you want to just add another sub-directory in the AddSystemModal folder? Or do you mean more than that?

ehelms:
Just a sub-directory

johnsonm325:
At this point, the inventory service is not working for us. Tests pass, but I still want to test it in dev environment. I won't merge until I know it's working as expected. Tests will pass even if some stuff isn't pointing to the correct place, so I'm not completely confident yet.

PR #151, opened 2019-06-21
Add tabs to add system modal

johnsonm325:
Can now select baselines and systems tabs in add system modal.


profiles/local-frontend-and-api.js
```diff
@@ -10,6 +10,7 @@ routes[`/beta/${SECTION}/${APP_ID}`] = { host: `http://localhost:${FRONTEND_PORT
 routes[`/${SECTION}/${APP_ID}`]      = { host: `http://localhost:${FRONTEND_PORT}` };
 routes[`/beta/apps/${APP_ID}`]       = { host: `http://localhost:${FRONTEND_PORT}` };
 routes[`/apps/${APP_ID}`]            = { host: `http://localhost:${FRONTEND_PORT}` };
+routes[`/api/system_baseline`]       = { host: `http://localhost:8085` };
```
beav: can you change `const API_PORT = 8080;` to be `DRIFT_PORT` and `BASELINE_PORT`, then use `BASELINE_PORT` here?

beav:
@johnsonm325 this change looks good. I am gonna ack, but we should consider adding some way to only expose this if users have a cookie set, or perhaps some boolean in the code that can be flipped easily per env.

PR #161, opened 2019-07-15
Switch to isBeta

johnsonm325:


beav:
change looks good. testing it now.

PR #162, opened 2019-07-17
Add Baselines page with populated list

johnsonm325:


beav:
@johnsonm325 can you get the code coverage percent to stay the same or go up? you could either add a test for the new stuff or remove the experimental cookie code that we aren't currently using, which should lower the total number of lines that need to be checked. It's already been checked into source control so we can re-add it whenever.

beav:
codecov didn't re-run but the percent is good: `25.75`

beav:
@johnsonm325 this tested correctly for me, thanks!

Can you make the baseline list page not render if the user is not on the beta site?  even if you just render "page not available at this time" or something.

PR #164, opened 2019-07-22
Update whats returned from the api response

johnsonm325:
To match what's returned by the api in system-baselines-backend/pr/10

### DIFF
--- a/src/SmartComponents/BaselinesTable/redux/__tests__/baselinesTableReducer.fixtures.js
+++ b/src/SmartComponents/BaselinesTable/redux/__tests__/baselinesTableReducer.fixtures.js
@@ -1,26 +1,20 @@
 /*eslint-disable camelcase*/
-const baselinesListPayload = ({
-    count: 2,
-    page: 1,
-    per_page: 50,
-    results: [
-        {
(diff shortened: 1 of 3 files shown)


src/SmartComponents/BaselinesTable/redux/__tests__/baselinesTableReducer.fixtures.js
```diff
@@ -1,26 +1,20 @@
...
-            created: '2019-02-18T13:30:00.000000Z',
-            display_name: 'micjohns baseline',
-            fact_count: 2,
-            id: 'abcd',
-            updated: '2019-05-19T15:00:00.000000Z'
-        }
-    ],
-    total: 2
-});
+const baselinesListPayload = ([
```
beav: do you want this as the payload, or the full payload with metadata and such?

johnsonm325: I think for what we use in the app, we just pass along this. After we make the api call, we only return the request.data.data. So the payload never had the meta data.

PR #167, opened 2019-08-14
Add ability to create an empty baseline

johnsonm325:
The baseline will be created with a name and then a blank table will be loaded into the UI for editing


beav:
looks ok to me. will test in a few

beav:
works for me in creating a new baseline

PR #168, opened 2019-08-20
Add buttons to create new facts and parent facts

johnsonm325:


beav:
```diff
+ Coverage   20.08%   20.5%   +0.41%     
```
:+1:

src/SmartComponents/BaselinesPage/EditBaseline/EditBaseline.js
```diff
@@ -11,15 +11,17 @@ import {
     TableTextInput
 } from '@patternfly/react-inline-edit-extension';
 import { Skeleton, SkeletonSize } from '@redhat-cloud-services/frontend-components';
+import { AddCircleOIcon } from '@patternfly/react-icons';
 
 import { baselinesPageActions } from '../redux';
 import { baselinesTableActions } from '../../BaselinesTable/redux';
-import api from '../../../api.js';
```
beav: what did this do previously?

beav:
i don't see anything egregious here code-wise. testing now

src/SmartComponents/BaselinesPage/EditBaseline/EditBaseline.js
johnsonm325: We were running a patch directly from the api. Now we call an action that calls the api so we can have the data in the payload.

beav:
There are some rough edges but I'd rather address them as bugs than in this PR.

ACK

PR #169, opened 2019-08-21
Add ability to add subfacts in UI

johnsonm325:


src/SmartComponents/BaselinesPage/EditBaseline/EditBaseline.js
```diff
@@ -241,6 +231,9 @@ class EditBaseline extends Component {
                 Object.keys(editedRowsBackup).forEach(key => {
                     rows[key].isEditing = false;
                 });
+                /*eslint-disable*/
+                console.log(rows);
+                /*eslint-enable*/
```
beav: looks like a console.log snuck in here

beav:
no-break of report: pass

beav:
the edit page still has some known rough spots but I was able to get through all of the scenarios with it

PR #170, opened 2019-08-22
Add CSV export to baselines list

johnsonm325:


src/SmartComponents/BaselinesPage/BaselinesKebab/redux/helpers.js
```diff
@@ -0,0 +1,52 @@
...
+    if (csv === null) {
+        return;
+    }
+
+    let filename = 'baseline-list-export-';
+    let today = new Date();
+    filename += today.getFullYear() + '-' + (today.getMonth() + 1) + '-' + today.getDate();
+    filename += '_';
+    filename += today.getHours() + ':' + today.getMinutes() + ':' + today.getSeconds();
+    filename += '.csv';
```
beav: it doesnt block merging of this PR, but is there a way to just get the datetime as an ISO8601 string and use that?

beav:
@johnsonm325 aside from the filename thing we discussed, this works for me.

I would be OK with treating the filename as a bug, up to you on if you want to fix it in this PR or not.

PR #171, opened 2019-08-23
use `rhel`, not `apps`

beav:
This changes `SECTION`. It also registers the left-hand nav which
makes things work in CI.


beav:
thanks @karelhala for the help with this

PR #172, opened 2019-08-23
Add csv export for baseline data

johnsonm325:


beav:
code looks ok, testing now

beav:
works for me

I thought it wasn't display subfacts at first, but I had to expand the menu and then it appeared

PR #175, opened 2019-08-30
Fix rendering issues on Edit Baseline Page

johnsonm325:
When typing values in text boxes for adding new facts and editing baseline the page would re-render, which became a problem with baselines that had a lot of facts. This PR keeps the page from re-rendering each time you type a letter.


beav:
code looks ok and it worked for me, thx!

PR #176, opened 2019-09-03
Add delete baseline

johnsonm325:


src/SmartComponents/BaselinesTable/BaselinesTable.js
```diff
@@ -13,7 +13,7 @@ class BaselinesTable extends Component {
         this.onSelect = this.onSelect.bind(this);
     }
 
-    async componentDidMount() {
+    componentDidMount() {
```
beav: I'm not famliar with what `async` did. What does its removal do?

johnsonm325: When I was talking with Karel a couple of weeks ago, he was saying that the async (asynchronous) wasn't necessary for this. I believe it's because the actual fetch is asynchronous.

ehelms:
What happens if a delete fails?

src/SmartComponents/BaselinesTable/BaselineTableKebab/BaselineTableKebab.js
```diff
@@ -14,11 +14,20 @@ class BaselineTableKebab extends Component {
...
         this.onKebabToggle = this.onKebabToggle.bind(this);
     }
 
-    findBaselineUUID() {
+    deleteBaselineData() {
+        const { fullBaselineListData, fetchBaselines, deleteBaseline, baselineRowData } = this.props;
+        let baselineId = baselinesTableHelpers.findBaselineId(baselineRowData, fullBaselineListData);
+
+        deleteBaseline(baselineId);
+        fetchBaselines();
```
ehelms: This performs a full fetch which may not have a performance impact. The alternative would be removing the deleted baseline (assuming success from server) from the data structure.

PR #181, opened 2019-09-10
Transition to new way of patching

johnsonm325:


ehelms:
@johnsonm325 could you summarize the change in how patching is handled? I think that would help me review.

johnsonm325:
This PR takes any changes made to the data from the api and creates a patch based on the diff of the original vs. the new data. An example would be:
```
var a = [
	{ name: 'a' },
	{ name: 'b' },
	{ name: 'c' },
]

var b = a.slice();
b.splice(1, 1);
b.push({ name: 'd' });

// Generate diff (ie JSON Patch) from a to b
var patch = jiff.diff(a, b);

// [{"op":"add","path":"/3","value":{"name":"d"}},{"op":"remove","path":"/1"}]
console.log(JSON.stringify(patch));
```
The patch is then sent along to the api call like `PATCH` to `/baselines/<uuid>` with the above `patch` as one of the parameters.

The jiff library is one based off of http://jsonpatchjs.com/. But, instead of having to write the patch by hand, this library will create it for you using .diff().

src/SmartComponents/BaselinesPage/EditBaseline/helpers.js
```diff
@@ -0,0 +1,37 @@
+import jiff from 'jiff';
```
ehelms: These functions look fairly straight forward in terms of expecting an input and returning a structured output. Given that, could this include some unit tests for this file?

PR #185, opened 2019-09-11
Updated time-date stamps to match UX requirements

johnsonm325:
Add systems table can not be updated from our app, but must be done from Inventory.
The last sync for baselines table (add system modal and baselines page table) and time in system comparison table have been updated


johnsonm325:
I'm going to have some failing tests here. I'll fix those.

PR #189, opened 2019-09-16
Fixes delay in fact filter when a lot of comparisons are being made

johnsonm325:
Bug: Add a lot of systems to a comparison (20+) then start typing in the fact filter bar. Each letter will fire off an action and will freeze up the system as it sorts the entire list.

Fix: There should be a 1000 millisecond delay after typing before the action fires off. If you continue typing, the timer will reset.

### DIFF
--- a/src/SmartComponents/DriftPage/SearchBar/SearchBar.js
+++ b/src/SmartComponents/DriftPage/SearchBar/SearchBar.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { TextInput } from '@patternfly/react-core';
 import PropTypes from 'prop-types';
+import _ from 'lodash';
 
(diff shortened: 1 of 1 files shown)


src/SmartComponents/DriftPage/SearchBar/SearchBar.js
```diff
@@ -10,13 +11,19 @@ export class SearchBar extends Component {
...
+
     render() {
         return (
             <React.Fragment>
                 <TextInput
+                    id="filterByFact"
                     placeholder="Filter by Fact"
-                    value={ this.props.factFilter }
-                    onChange={ this.props.changeFactFilter }
+                    onChange={ _.debounce(this.props.changeFactFilter, 1000) }
```
ehelms: This amount of time, for me, made the application feel slow. I lowered it to 250 locally and things felt good while having a delay.

PR #190, opened 2019-09-23
use baselines and systems when creating CSV header

beav:
Previously, only systems were used when creating the CSV file header.
Instead, use both systems and baselines if available.

### DIFF
--- a/src/SmartComponents/modules/reducers.js
+++ b/src/SmartComponents/modules/reducers.js
@@ -145,7 +145,7 @@ export function compareReducer(state = initialState, action) {
                 totalFacts: filteredFacts.length
             };
         case `${types.EXPORT_TO_CSV}`:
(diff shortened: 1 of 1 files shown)


src/SmartComponents/modules/reducers.js
```diff
@@ -145,7 +145,7 @@ export function compareReducer(state = initialState, action) {
                 totalFacts: filteredFacts.length
             };
         case `${types.EXPORT_TO_CSV}`:
-            reducerHelpers.downloadCSV(state.sortedFilteredFacts, state.systems);
+            reducerHelpers.downloadCSV(state.sortedFilteredFacts, [ ...state.systems, ...state.baselines ]);
```
ehelms: If we swap the order of these the CSV will match how its rendered in the UI (baselines first, then systems)

PR #193, opened 2019-09-26
Fix system names unaligned to checkbox

ehelms:


beav:
I get the same behavior as before with this patch

image

PR #195, opened 2019-09-27
Add Breadcrumb to baselines page

johnsonm325:


johnsonm325:
@ehelms One of theses tests fails and I can not figure out how to make it pass. I'm hoping you can help with this.

src/Routes.js
```diff
@@ -27,6 +27,9 @@ export const Routes = () => {
             { window.insights.chrome.isBeta() ?
                 <InsightsRoute exact path='/baselines' component={ BaselinesPage } /> : null
             }
+            { window.insights.chrome.isBeta() ?
+                <InsightsRoute path='/baselines/:id' component={ BaselinesPage } /> : null
```
ehelms: This feels like we are now overloading the BaselinesPage component and that if I go to this specific URL it should just render the BaselineEditPage

ehelms:
The test data looks exactly the same from eyeball test. I wonder if its the deprecation warning thrown by moment that is failing the test:

```
    Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
```

src/Routes.js
johnsonm325: That's fair. I'm not sure at this moment how much of a change this is going to require. Theoretically it should be an easy switch, but I'll need to work through it.

johnsonm325:
Well, I had 2 failing tests, but then I updated it to add the .selected = true and one of them passed.

ehelms:
I think this is the cuase of the failing test:

https://github.com/facebook/jest/issues/5998

johnsonm325:
> I think this is the cuase of the failing test:
> 
> [facebook/jest#5998](https://github.com/facebook/jest/issues/5998)

I tried this already, and that didn't seem to help. Maybe I did it incorrectly?

src/SmartComponents/BaselinesPage/CreateBaseline/CreateBaseline.js
```diff
@@ -28,14 +29,16 @@ class CreateBaseline extends Component {
         };
     }
 
-    submitBaselineName() {
+    async submitBaselineName() {
         const { baselineName } = this.state;
-        const { createBaseline, toggleCreateBaseline } = this.props;
+        const { createBaseline, toggleCreateBaseline/*, fetchBaselineData*/ } = this.props;
```
ehelms: Inline commented code here

src/SmartComponents/BaselinesPage/CreateBaseline/CreateBaseline.js
```diff
@@ -28,14 +29,16 @@ class CreateBaseline extends Component {
...
         /*eslint-disable camelcase*/
         let newBaselineObject = { display_name: baselineName, baseline_facts: []};
         /*eslint-enable camelcase*/
 
-        createBaseline(newBaselineObject);
+        await createBaseline(newBaselineObject);
+        //fetchBaselineData(baselineData.id);
```
ehelms: Inline commented code here

ehelms:
I think if we set https://github.com/RedHatInsights/drift-frontend/pull/195/files#diff-ea5577af28ec2c480dfae0e76a6fbd60R5 to true by default we will not see the empty state flash.

PR #197, opened 2019-09-30
Add search bar to Baselines table

ehelms:
This change moves the empty state logic and fetching logic up to the
page itself and out of the table. This simplifies the table and lets
the page control the flow. Given the Page component is where all of
the parts are composed it necessitates that the logic of what to
present is controlled here.


beav:
works for me and code looks good

PR #199, opened 2019-10-04
Fix beta bits landing in ci-stable

ehelms:


beav:
i'm having some trouble understanding this change. What is the fix doing?

ehelms:
When we build the UI code, it takes into consideration the environment being built for and hard codes within the built code references to either the beta or production environment. When a PR gets merged to master, I want to push those changes to both CI and CI beta directly. In order to do that, we have to build it once with the beta flag and once without the beta flag and push each respectively to ci-beta or ci-stable.

.travis/custom_release.sh
```diff
@@ -6,15 +6,16 @@ if [ "${TRAVIS_BRANCH}" = "master" ]; then
     echo
     echo
     echo "PUSHING ci-beta"
-    .travis/release.sh "ci-beta"
-    rm -rf dist/.git
+    BUILD_BETA=true .travis/release.sh "ci-beta"
```
ehelms: I need to check and see if this does an npm build or if thats done prior to this which would make this inefficient.

ehelms:
This is now updated. It should be default build for master a beta based build. Then wipe away the build, and re-build for stable. The beta should get pushed to ci-beta and the stable to ci-stable

config/webpack.common.js
```diff
@@ -17,8 +17,12 @@ const betaBranch =
     gitBranch === 'master' ||
     gitBranch === 'qa-beta' ||
     gitBranch === 'prod-beta';
+const buildStable = process.env.BUILD_STABLE === 'true';
 
-if ((process.env.NODE_ENV === 'production' && betaBranch) || process.env.BUILD_BETA === 'true') {
+const buildBeta = (betaBranch || process.env.BUILD_BETA === 'true') && !buildStable;
+
+if (process.env.NODE_ENV === 'production' && buildBeta) {
+    console.log('building beta');
```
beav: just confirming, this log statement is intentional and not debug, correct?

ehelms: Good catch, dropped.

PR #204, opened 2019-10-04
Fix baselineListLoading being pulled from redux state

ehelms:


src/SmartComponents/BaselinesTable/BaselinesTable.js
```diff
@@ -160,6 +160,8 @@ class BaselinesTable extends Component {
                 </Table>;
             }
         } else {
+            /*eslint-disable*/
+            console.log("LOADING");
```
beav: loading please wait

PR #207, opened 2019-10-07
Auto push to QA beta and stable

ehelms:


beav:
@ehelms do we need to also push services to QA automatically? otherwise they might not match

ehelms:
I tend to think the UI is more fluid and changes more rapidly (and in less response to) the backend. The backend gets pushed when we push to stable which I think is fine for now. That's my way of saying I think we should try this flow and push to stable more often (and perhaps automatically in time).

beav:
@ehelms works for me

PR #209, opened 2019-10-07
Fixed bugs with editing facts

johnsonm325:
Adding new subfact replaces existing subfact
Updating sub-fact value results in all other sub-facts disappearing
Actions on facts/categories is not matching after fact/sub-fact editing
Editing category deletes sub-facts
Delete facts functionality doesn't work properly


beav:
:heavy_check_mark: Adding new subfact replaces existing subfact
:heavy_check_mark:  Updating sub-fact value results in all other sub-facts disappearing
Actions on facts/categories is not matching after fact/sub-fact editing
:heavy_check_mark: Editing category deletes sub-facts
:heavy_check_mark: Delete facts functionality doesn't work properly

PR #210, opened 2019-10-08
Resets filters when 1 or no systems/baselines remain

johnsonm325:
Fixes Reset view filter to 'all' when only one system remains in the list
Fixes When removing all systems/baselines by 'Clear all comparisons' option, all filters must be reset to 'selected'


beav:
:heavy_check_mark: Fixes Reset view filter to 'all' when only one system remains in the list
:heavy_check_mark:  Fixes When removing all systems/baselines by 'Clear all comparisons' option, all filters must be reset to 'selected'

both bugs tested correctly with this PR

PR #214, opened 2019-10-09
Add clone options for creating baseline

johnsonm325:


beav:
code looks OK to me, given that it's stub code to show something in the UI for a demo.

will test in 5-10 min

src/SmartComponents/BaselinesPage/CreateBaselineModal/CreateBaselineModal.js
```diff
@@ -14,12 +14,27 @@ class CreateBaselineModal extends Component {
         this.submitBaselineName = this.submitBaselineName.bind(this);
 
         this.state = {
-            baselineName: ''
+            baselineName: '',
+            check1: false,
+            check2: false,
+            check3: false
```
ehelms: I know this is a kinda POC, but if its also the starter work for this change, I would ask we consider giving these checks names that are meaningful.

beav:
can you squash this into one commit?

src/SmartComponents/BaselinesPage/CreateBaselineModal/CreateBaselineModal.js
```diff
@@ -41,10 +70,67 @@ class CreateBaselineModal extends Component {
...
-        const { baselineName } = this.state;
+    renderRadioButtons() {
+        const { fromScratchChecked, copyBaselineChecked, copySystemChecked } = this.state;
+
+        return (<React.Fragment>
+            <Radio
+                isChecked={ fromScratchChecked }
+                id='create baseline'
+                label='Create baseline from scratch'
+                name='radio-1'
```
beav: what's the difference between name and id? does name have to be radio-1, radio-2, radio-3?

johnsonm325: I built it like the example from patternfly, but honestly, I think we could trash the name field. What do you think?

```
import React from 'react';
import { Radio } from '@patternfly/react-core';

class ControlledRadio extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      check1: false,
    };

    this.handleChange = (_, event) => {
      const { value } = event.currentTarget;
      this.setState({ [value]: true });
    };
  }

  render() {
    return (
      <React.Fragment>
        <Radio
          isChecked={this.state.check1}
          name="radio-1"
          onChange={this.handleChange}
          label="Controlled radio"
          id="radio-controlled"
          value="check1"
        />
      </React.Fragment>
    );
  }
}
```

beav:
the multi-select needs to be removed from baseline and system select:

image

beav:
also the "copy from existing system" modal only shows one line since it has a search bar. I don't have an issue with that but you may want to run it by katie

src/SmartComponents/BaselinesPage/CreateBaselineModal/CreateBaselineModal.js
```diff
@@ -14,22 +16,49 @@ class CreateBaselineModal extends Component {
...
 
     async submitBaselineName() {
-        const { baselineName } = this.state;
-        const { createBaseline, toggleCreateBaselineModal } = this.props;
+        const { baselineName, fromScratchChecked } = this.state;
+        const { createBaseline, toggleCreateBaselineModal, selectedBaselineIds, entities } = this.props;
+        let newBaselineObject;
+
         /*eslint-disable camelcase*/
-        let newBaselineObject = { display_name: baselineName, baseline_facts: []};
```
ehelms: If you kept this but dropped the last param then the if-else would be a bit cleaner I think:

```
        if (fromScratchChecked) {
            newBaselineObject['baseline_facts'] =  [];
            await createBaseline(newBaselineObject);
        } else if (selectedBaselineIds.length === 1) {
            await createBaseline(newBaselineObject, selectedBaselineIds[0]);
        } else if (entities.selectedSystemIds.length === 1) {
            newBaselineObject['inventory_uuid'] = entities.selectedSystemIds[0];
            await createBaseline(newBaselineObject);
        }
```

src/SmartComponents/BaselinesTable/BaselinesTable.js
```diff
@@ -18,7 +18,10 @@ import { baselinesTableActions } from './redux';
 class BaselinesTable extends Component {
     constructor(props) {
         super(props);
-        this.onSelect = this.onSelect.bind(this);
+
+        this.state = {
+            createBaselineModal: this.props.createBaselineModal
+        };
```
ehelms: Is there an advatange to doing this here for in `mapStateToProps` ?

src/SmartComponents/BaselinesTable/BaselinesTable.js
```diff
@@ -91,22 +107,31 @@ class BaselinesTable extends Component {
         return table;
     }
 
+    createModalRows() {
```
ehelms: What makes a modal row a modal row vs. in the normal table?

src/SmartComponents/BaselinesTable/redux/baselinesTableReducer.js
```diff
@@ -61,6 +61,33 @@ export function baselinesTableReducer(state = initialState, action) {
...
+
+            state.baselineTableData.map(row => {
+                if (action.payload.id.includes(row[0])) {
+                    row.selected = action.payload.isSelected;
+                } else {
+                    row.selected = false;
+                }
+
+                newBaselineTableData.push(row);
+            });
```
ehelms: I think if we re-factored the logic in the SELECT_BASELINE reducer to some helpers, we could use one action to handle both cases. Not required for this change and I might take a crack at doing that.

src/SmartComponents/BaselinesTable/BaselinesTable.js
johnsonm325: At the moment, modalRows have the checkboxes and don't have the kebab. The normal table has kebab but no checkboxes.

src/SmartComponents/BaselinesTable/redux/baselinesTableReducer.js
johnsonm325: Agree, but definitely want to tackle this later. I'll make a card for it so we ensure it gets addressed.

src/SmartComponents/BaselinesTable/BaselinesTable.js
johnsonm325: modalRows have checkboxes and no kebabs

beav:
@ehelms any further comment?

PR #215, opened 2019-10-09
Allow persistent baseline select in modal

ehelms:
This reworks the select logic in order to fix a bug where selecting a baseline in the system modal would not persist. Before this fix if you clicked on the System tab or performed a baseline search, any selected baselines would lose their selections. Now you should be able to switch back and forth or perform search without losing any selection.

This has failing tests and some code cleanup yet to be done. @johnsonm325 wanted you to see the changes I was proposing to the logic as this moves some of it out of the component and into redux.

### DIFF
--- a/src/SmartComponents/BaselinesTable/BaselinesTable.js
+++ b/src/SmartComponents/BaselinesTable/BaselinesTable.js
@@ -30,19 +30,17 @@ class BaselinesTable extends Component {
 
     onSelect(event, isSelected, rowId) {
         const { baselineTableData, selectBaseline } = this.props;
-        let rows;
+        let ids;
 
         if (rowId === -1) {
-            rows = baselineTableData.map(oneRow => {
-                oneRow.selected = isSelected;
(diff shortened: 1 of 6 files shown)


src/SmartComponents/BaselinesTable/BaselinesTable.js
```diff
@@ -30,19 +30,17 @@ class BaselinesTable extends Component {
...
+                return item[0];
             });
         } else {
-            rows = [ ...baselineTableData ];
-            rows[rowId].selected = isSelected;
+            ids = [ baselineTableData[rowId][0] ];
         }
 
-        selectBaseline(rows);
+        selectBaseline(ids, isSelected, baselineTableData);
```
johnsonm325: Here you're passing 3 arguments to selectBaseline, but in the actions, you are only accepting 2.

ehelms:
Thanks for catching those old design leftovers. Updated now

PR #216, opened 2019-10-11
Add validation reporting when creating a baseline

ehelms:


beav:
@ehelms a couple of tests are failing but I think they just need their expected result updated for the PR's changes

beav:
@ehelms this scenarios fails:

* try to create a baseline with same name as existing, get an error (correct)
* click X, go back to the baseline list, hit "create baseline" again, old error is still there

image

beav:
scenario above still fails with updated code

PR #217, opened 2019-10-11
Makes fact filter on comparison table persistent

johnsonm325:
Fixes:
Filter value is kept when adding a baseline to a comparison but not showing in the box


beav:
code change looks ok to me, testing now

beav:
works for me

PR #219, opened 2019-10-15
link radio buttons together to disable multiselect

beav:
The radio buttons did not have a shared name, so users were able to
select more than one at a time. This commit adds a name.


beav:
I had trouble reproducing this issue in my local env. However I think this will fix the issue in CI.

johnsonm325:
Well, if it works in ci then we'll know why we should have kept the names in there instead of deleting them like I did on the previous PR.

PR #221, opened 2019-10-17
Temporarily block submit baseline button when no data is selected

johnsonm325:
Also increased the size of the modal to show more than 1 system


src/App.scss
```diff
@@ -189,7 +189,7 @@
   .create-baseline-modal {
     width: initial;
     max-width: 1200px;
-    max-height: 600px;
+    max-height: 700px;
```
beav: thanks!

PR #222, opened 2019-10-18
Drop beta feature checks for baselines

ehelms:


src/Routes.js
```diff
@@ -25,12 +25,8 @@ InsightsRoute.propTypes = {
 export const Routes = () => {
     return (
         <Switch>
-            { window.insights.chrome.isBeta() ?
-                <InsightsRoute exact path='/baselines' component={ BaselinesPage } /> : null
-            }
-            { window.insights.chrome.isBeta() ?
-                <InsightsRoute path='/baselines/:id' component={ EditBaseline } /> : null
-            }
+            <InsightsRoute exact path='/baselines' component={ BaselinesPage } /> : null
```
beav: looks like you have both sides of the if/else here "`: null`"

ehelms: You are right, fixed it up

PR #225, opened 2019-10-22
Make UI updates based on review with UX

johnsonm325:


johnsonm325:
Added a couple of UI edits to this.

PR #227, opened 2019-10-23
Clear baselines when changing between Sys Compare and Baselines menus

johnsonm325:


ehelms:
I think this solution will work for todays landscape but it may not always work in the future as it makes the assumption that the only place a user can go is to the baselines page. If I look ahead to some work we are considering around setting fact prioritization, and if that introduces a new page a user can get to this might show back up. Is it possible to engineer the solution so its a bit more robust?

johnsonm325:
We could have it clear selected baselines when the drift page is rendered.

PR #229, opened 2019-10-29
Re-factor the Toolbar into a component and replace the search bar

ehelms:
This moves the toolbar logic into it's own component that is
configurable through parameters for it's various use cases. The
tool bar is centralized into the BaselinesTable letting it control
the entirety of the fetching logic.


src/App.scss
```diff
@@ -28,6 +28,7 @@
   }
 
   .pf-c-card > :last-child.pf-c-card__body {
+    padding-top: 0;
```
johnsonm325: It looks like this has caused the `Add fact` button on the EditBaseline.js page to touch the top of the card. May just need to add margin-top to the `display-margin` class, but I'm bad at CSS and you may know a better method.

ehelms: Fixed

src/App.scss
```diff
@@ -64,11 +65,6 @@
   }
 
   .drift-toolbar {
-    padding: 20px;
-    padding-top: 0px;
-  }
-
-  .drift-modal-toolbar {
     padding: 16px 32px;
```
johnsonm325: This is throwing off the padding for the toolbar in the System Comparison table. You can compare it's state in this PR to ci and see the difference.

ehelms:
@johnsonm325 I think this is ready for final review if oyu have a chance. I can then proceed with adding sort functionality.

PR #230, opened 2019-10-30
Add multi-select and delete to baselines table

johnsonm325:


src/SmartComponents/BaselinesPage/BaselinesKebab/BaselinesKebab.js
```diff
@@ -1,78 +1,86 @@
...
+            dropdownItems = [
+                <DropdownItem
+                    key="multi-delete"
+                    component="button"
+                    onClick={ this.toggleModalOpened }
+                    isDisabled
+                >
+                    Delete baselines
+                </DropdownItem>
+            ];
```
ehelms: You can simplify this by dropping the if/else and doing:

```
           dropdownItems = [
                <DropdownItem
                    key="multi-delete"
                    component="button"
                    onClick={ this.toggleModalOpened }
                    isDisabled={ selectedBaselineIds.length > 0 }
                >
                    Delete baselines
                </DropdownItem>
            ];
```

Note how the `isDisabled` takes a value and can be set based on a condition reducing the repetition.

ehelms:
Personally, I would like to get the Toolbar refactor PR (https://github.com/RedHatInsights/drift-frontend/pull/229) in and then rebasing this change on top of that so that the design is congruent.

src/SmartComponents/BaselinesPage/BaselinesKebab/BaselinesKebab.js
johnsonm325: I didn't know that! I can fix a number of things with this. It makes sense how that would work though.

src/SmartComponents/BaselinesPage/CreateBaselineModal/CreateBaselineModal.js
```diff
@@ -108,7 +109,8 @@ class CreateBaselineModal extends Component {
     renderCopyBaseline() {
         return (<React.Fragment>
             <b>Select baseline to copy from</b>
-            <BaselinesTable createBaselineModal='true' />
+            { /*<BaselinesTable createBaselineModal={ true } hasSelect={ true } />*/ }
```
ehelms: I assume this needs removing

src/SmartComponents/BaselinesPage/DeleteBaselinesModal/DeleteBaselinesModal.js
```diff
@@ -0,0 +1,98 @@
...
+        /*eslint-disable camelcase*/
+        if (baselineId) {
+            apiBody = { data: { baseline_ids: [ baselineId ]}};
+        } else {
+            apiBody = { data: { baseline_ids: selectedBaselineIds }};
+        }
+        /*eslint-enable camelcase*/
+
+        this.toggleModal();
+        clearSelectedBaselines();
```
ehelms: If we clear and then the delete fails, I feel like we'd be in a bad state.

src/SmartComponents/BaselinesPage/DeleteBaselinesModal/DeleteBaselinesModal.js
```diff
@@ -0,0 +1,98 @@
...
+        );
+    }
+}
+
+DeleteBaselinesModal.propTypes = {
+    modalOpened: PropTypes.bool,
+    clearSelectedBaselines: PropTypes.func,
+    selectedBaselineIds: PropTypes.array,
+    deleteSelectedBaselines: PropTypes.func,
+    fetchBaselines: PropTypes.func
```
ehelms: This modal ends up having to know a lot about the inner workings of the baselines page and table. Could it be simplified to be passed a function that handles all of the deleting, clearing, knowledge of what is selected and then fetching?

src/SmartComponents/BaselinesTable/BaselinesTable.js
```diff
@@ -39,6 +38,14 @@ class BaselinesTable extends Component {
         this.fetchBaselines();
     }
 
+    componentDidUpdate(prevProps) {
+        const { fetchBaselines } = this.props;
+
+        if (this.props.createBaselineModalOpened === false && prevProps.createBaselineModalOpened === true) {
+            fetchBaselines();
+        }
```
ehelms: What case is this handling? I've not seen us use `componentDidUpdate` before.

src/SmartComponents/BaselinesPage/DeleteBaselinesModal/DeleteBaselinesModal.js
johnsonm325: I think the issue with not deselecting is that it would still be full of the uuids of the deleted baselines. I think we have to clear it, but maybe we check the success status of the delete first.

johnsonm325: The only "bad state" would be that the person would have to reselect all of the baselines.

src/SmartComponents/BaselinesTable/BaselinesTable.js
johnsonm325: Found this bug. If you try to clone a baseline from another baseline by selecting a baseline, and then exiting the modal without actually cloning, then the baseline would be selected in the baseline table. So this checks to see if the previous state of the modal was opened and that it is now closed. If so, then it does another fresh fetch and deselects the baselines.

PR #231, opened 2019-10-30
Add error output validation for edit baseline name and facts

ehelms:


ehelms:
This shouldn't disable category on 'Add Fact'

ehelms:
I have made a decent round of updates to this to more accurately send patch requests and get results for add vs. edit states. If you try to 'Add fact' a fact with an existing name and value, you should now see the server report a 500 error. This shows the validation in action but is obviously a backend bug.

johnsonm325:
I'm getting the error message on both the name box and value box. Is that expected?
image
Then, when I tried to change the name right there in the box, I get the following error:
image

johnsonm325:
Another bug I found. If you try to edit a category, the "This is a category" box is not disabled. We want it so once you've created a category and then try to edit it, you can't switch that on/off again.

beav:
:heart:

PR #232, opened 2019-11-05
Add sort to baselines table

ehelms:


beav:
@ehelms the sort works well for me, but the "last updated" column header bounces inward and then back out again once the results are returned.

I'm happy to pull this PR in as-is. I don't know if the column movement is something that can be fixed in this PR or if it should be a new issue.

PR #237, opened 2019-11-19
Fix ghost selection comparison when removing a system

ehelms:
If all systems and baselines were removed in a comparison, followed
by a comparison of just baselines the internal saved set of system
ids would not reset.


beav:
:ghost:

PR #238, opened 2019-12-10
Add multi-fact delete functionality

johnsonm325:
Tests written


beav:
after deleting all facts via the "select all" box, the box is still checked:

image

beav:
I hit one small issue but looks good to me otherwise.

johnsonm325:
Ok, should be fixed.

src/SmartComponents/BaselinesPage/EditBaseline/DeleteFactModal/DeleteFactModal.js
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    buildMessage = (categoryMessage, factMessage) => {
+        let numFactsMessage;
+        let deleteCatMessage;
+        let message;
+
+        if (categoryMessage && factMessage) {
+            numFactsMessage = 'You have selected ' + categoryMessage + ' and ' + factMessage + ' to be deleted.';
+            deleteCatMessage = 'Deleting a category will delete all facts within the category.';
```
ehelms: You define this on line 45 as well. So it can be moved out of the if-else block to the top of the function.

src/SmartComponents/BaselinesPage/EditBaseline/DeleteFactModal/DeleteFactModal.js
```diff
@@ -0,0 +1,174 @@
...
+            numFactsMessage = 'You have selected ' + factMessage + ' to be deleted.';
+            message = <div>
+                <span className='display-block'>{ numFactsMessage }</span>
+            </div>;
+        }
+
+        return message;
+    }
+
+    countFacts = () => {
```
ehelms: For me, this is an unituitive name for what it returns. This function both counts facts and returns a message about them. The counting could be moved out to a helper for easier testing and this function focus on generating the message inside the component.

src/SmartComponents/BaselinesPage/EditBaseline/DeleteFactModal/DeleteFactModal.js
```diff
@@ -0,0 +1,174 @@
...
+
+        return message;
+    }
+
+    render() {
+        const { modalOpened } = this.state;
+        const { deleteFact, isCategory } = this.props;
+        let message;
+
+        if (deleteFact) {
```
ehelms: What is the deleteFact property? Is this the use case to handle the kabob issuing a delete?

src/SmartComponents/BaselinesPage/EditBaseline/redux/reducers.js
```diff
@@ -93,6 +92,36 @@ export function editBaselineReducer(state = initialState, action) {
...
+                        if (action.payload.ids.includes(subFact[0])) {
+                            subFact.selected = action.payload.isSelected;
+                        }
+                    });
+                }
+            });
+
+            newEditBaselineTableData.forEach(function(row) {
+                if (!row.selected) {
+                    isAllSelected = false;
```
ehelms: You could put a `break;` after this to exit the loop early since you'll have hit the condition.

src/SmartComponents/BaselinesPage/EditBaseline/redux/reducers.js
```diff
@@ -93,6 +92,36 @@ export function editBaselineReducer(state = initialState, action) {
                 isCategory: action.payload.isCategory,
                 isSubFact: action.payload.isSubFact
             };
+        case `${types.SELECT_FACT}`: {
+            newEditBaselineTableData = [ ...state.editBaselineTableData ];
+            let isAllSelected = true;
+
+            newEditBaselineTableData.map(row => {
+                if (action.payload.ids.includes(row[0])) {
```
ehelms: What are `row[0]` and `row[2]` ? Casting them to variables at the top of this loop would help in understanding what is happening here,e.g.:

```
let id = row[0]
let ?? = row[2]
```

src/SmartComponents/BaselinesPage/EditBaseline/redux/reducers.js
```diff
@@ -93,6 +92,36 @@ export function editBaselineReducer(state = initialState, action) {
...
+                if (Array.isArray(row[2])) {
+                    row[2].map(subFact => {
+                        if (action.payload.ids.includes(subFact[0])) {
+                            subFact.selected = action.payload.isSelected;
+                        }
+                    });
+                }
+            });
+
+            newEditBaselineTableData.forEach(function(row) {
```
ehelms: I saw a helper you added similar to this -- https://github.com/RedHatInsights/drift-frontend/pull/238/files#diff-f5121fc597b048d54fc72bd79b451ed1R331

src/SmartComponents/BaselinesPage/EditBaseline/helpers.js
```diff
@@ -272,25 +302,57 @@ function filterBaselineData(baselineData, expandedRows) {
...
+}
+
+function findIsSelected(factData, id, isSelected) {
+    let allSubFactsSelected = true;
+
+    for (let i = 0; i < factData.length; i++) {
+        if (factData[i][0] === id && isSelected && !factData[i].selected) {
+            allSubFactsSelected = true;
+        } else if (!factData[i].selected) {
+            return false;
```
ehelms: The function makes me think it will always return an array of selected values. This return here exits the for loop and would give me back a boolean breaking my expectation.

src/SmartComponents/BaselinesPage/EditBaseline/helpers.js
```diff
@@ -140,6 +140,42 @@ function makeDeleteSubFactPatch(factToDelete, parentFact, originalAPIBody) {
     return newAPIBody;
 }
 
+function makeDeleteFactsPatch(factsToDelete, originalAPIBody) {
+    let patchBody = [];
+
+    factsToDelete.forEach(function(fact, index) {
+        if (!fact.selected) {
+            if (Array.isArray(fact[2])) {
```
ehelms: What does `fact[2] represent? Casting to a variable with a name that identifies what it is would help here.

src/SmartComponents/BaselinesPage/EditBaseline/helpers.js
```diff
@@ -140,6 +140,42 @@ function makeDeleteSubFactPatch(factToDelete, parentFact, originalAPIBody) {
...
+                };
+
+                patchBody.push(newFact);
+            } else {
+                patchBody.push(originalAPIBody.baseline_facts[index]);
+            }
+        }
+    });
+
+    let patch = makePatchBody(patchBody, originalAPIBody.baseline_facts);
```
ehelms: It can be a bit confusing naming wise to see a patchBody being sent to something that makes a patch body...

src/SmartComponents/BaselinesPage/EditBaseline/helpers.js
```diff
@@ -3,13 +3,13 @@ import jiff from 'jiff';
...
 /*eslint-disable react/prop-types*/
-function renderKebab({ factName, factValue, fact, isCategory, isSubFact } = {}) {
+function renderKebab({ factName, factValue, factData, isCategory, isSubFact } = {}) {
     return (
         <td>
             <FactKebab
                 factName={ factName }
                 factValue={ factValue }
-                fact={ fact }
+                fact={ factData }
```
ehelms: What is `factData` at this point? We pass in fact name and fact value, which is the fact data

src/SmartComponents/BaselinesPage/EditBaseline/EditBaselineKebab/EditBaselineKebab.js
```diff
@@ -0,0 +1,99 @@
...
+            this.setState({
+                modalOpened: !modalOpened,
+                kebabOpened: false
+            });
+        };
+    }
+
+    isFactSelected = () => {
+        const { editBaselineTableData } = this.props;
+        let isDisabled = true;
```
ehelms: I think this is mixing concerns. The function is all about figuring out if a fact is selected. This function is used by another chunk of the code to determine whether it disables something. But this function, shouldn't care about disabling or not, but rather is a fact selected. So I think changing this to `isSelected` would be cleaner.

src/SmartComponents/BaselinesPage/EditBaseline/EditBaselineKebab/EditBaselineKebab.js
```diff
@@ -0,0 +1,99 @@
...
+
+        editBaselineTableData.forEach((fact) => {
+            if (fact.selected) {
+                isDisabled = false;
+            }
+
+            if (Array.isArray(fact[2]) && fact[2].length > 0) {
```
ehelms: I'm starting to wonder if we should introduce some constants to make things more eadable across all these components that deal with arrays. Either:

```
BASELINE_SUB_FACTS = 2

if (Array.isArray(fact[BASELINE_SUB_FACTS]) && fact[BASELINE_SUB_FACTS].length > 0) {
```

or with a function:

```
isCategory(fact) {
  let subfacts = fact[2];
  if (Array.isArray(subfacts) && subfacts.length > 0) {
    return true
  } else 
   return false;
  }
}

baseline_sub_facts(fact) {
    return fact[2];
}

if (isCategory(fact)) {
  baseline_sub_facts(fact).forEach(subFact) => {
```

src/SmartComponents/BaselinesPage/EditBaseline/redux/reducers.js
johnsonm325: Can't break or continue in a forEach loop. Can change this to for loop instead?

src/SmartComponents/BaselinesPage/EditBaseline/DeleteFactModal/DeleteFactModal.js
johnsonm325: This is a function passed from the FactKebab file that helps keeps the modal as dumb as possible about the contents of the fact. https://github.com/RedHatInsights/drift-frontend/blob/4237ff92a1346ba9683e58aebefeda287e4667a3/src/SmartComponents/BaselinesPage/EditBaseline/FactKebab/FactKebab.js#L119

src/SmartComponents/BaselinesPage/EditBaseline/helpers.js
johnsonm325: It would be factName and factValue unless it is a category, and then it would contain factName and the subFacts.

src/SmartComponents/BaselinesPage/EditBaseline/DeleteFactModal/helpers.js
```diff
@@ -0,0 +1,26 @@
+function countFacts(editBaselineTableData) {
+    let categories = 0;
+    let facts = 0;
+
+    editBaselineTableData.forEach(function(fact) {
+        if (fact.selected) {
+            if (Array.isArray(fact[2])) {
```
ehelms: This can use the new helper functions

src/SmartComponents/BaselinesPage/EditBaseline/DeleteFactModal/helpers.js
```diff
@@ -0,0 +1,26 @@
...
+                categories += 1;
+            } else {
+                facts += 1;
+            }
+        } else if (Array.isArray(fact[2])) {
+            fact[2].forEach(function(subFact) {
```
ehelms: This can use the new helper functions

src/SmartComponents/BaselinesPage/EditBaseline/EditBaseline.js
```diff
@@ -97,44 +104,123 @@ class EditBaseline extends Component {
...
+        selectFact(facts, isSelected);
+    }
+
+    renderCheckbox = (fact) => {
+        const { selectAll } = this.props;
+        let id;
+
+        if (typeof(fact[2]) === 'string') {
+            id = 'fact-' + fact[0];
+        } else if (Array.isArray(fact[2]) && fact[2].length > -1) {
```
ehelms: This can use the new helper functions

src/SmartComponents/BaselinesPage/EditBaseline/EditBaseline.js
```diff
@@ -97,44 +104,123 @@ class EditBaseline extends Component {
...
+        let factData = baselineData.baseline_facts.find((baselineFact) => {
+            return baselineFact.name === fact[1];
+        });
+
+        row.push(<td
+            className={ expandedRows.includes(fact[1]) ? 'pf-c-table__check nested-fact' : 'pf-c-table__check' }>
+            { this.renderCheckbox(fact) }
+        </td>);
+
+        if (Array.isArray(fact[2]) && fact[2].length > -1) {
```
ehelms: This can use the new helper functions

src/SmartComponents/BaselinesPage/EditBaseline/EditBaseline.js
```diff
@@ -97,44 +104,123 @@ class EditBaseline extends Component {
...
+                    rows.push(<tr key={ subFact[1] }>{ row }</tr>);
                 });
             }
         } else {
-            row.push(<td>{ fact.name }</td>);
-            row.push(<td>{ fact.value }</td>);
-            row.push(editBaselineHelpers.renderKebab({ factName: fact.name, factValue: fact.value, fact }));
-            rows.push(<tr>{ row }</tr>);
+            row.push(<td>{ fact[1] }</td>);
+            row.push(<td>{ fact[2] }</td>);
```
ehelms: This can use the new helper functions

ehelms:
One of the general ideas for my suggestion around the identification helpers is to make it easier to read and understand what is being referenced rather than array indexes since its very easy to lose track of them, e.g. facts[1], facts[2]

src/SmartComponents/BaselinesPage/EditBaseline/EditBaseline.js
johnsonm325: It can't because the helper function looks for categories that have subFacts. This is checking for a category that has at least an empty array. I could create two different helper functions, isCategory and isCategoryWithFacts.

johnsonm325:
This should be ready for a final review.

ehelms:
I am noticing that if I select a category, it selects all facts and categories.

ehelms:
Also noticing if I try to create a category with this PR it creates a regular fact with no value.

johnsonm325:
I'm not noticing the first part. I selected a category and it only selected that category and it's sub-facts:
image

I totally see the second issue though.

PR #239, opened 2019-12-17
Add profiles in time

johnsonm325:


beav:
patch works for me as expected

beav:
I tested this against QA services as well which have the older code. It seems to be backwards compatible to me.

PR #242, opened 2020-01-07
Fix padding on system compare page

johnsonm325:


johnsonm325:
This is broken at the moment, so don't merge

PR #247, opened 2020-01-22
Visual refresh for system/baseline header and icons

ehelms:


katierik:
Looks overall really great. Have a few UI notes (if these are not readable/understandable let me know)
Xnip2020-01-22_14-33-02
Xnip2020-01-22_14-32-29

Variables for indicating stale systems:

- pf-global-warning-color--200 (stale)

- pf-global-danger-color--100 (scheduled to be deleted)

Background color for system:

- pf-color-black-0200

^ Both of these are from the design files, so hopefully they are correct

ehelms:
@katierik are you ok with the placement of the 'x' ? I was finding it difficult to align it on the same line as the hostname reliably for both long hostnames and changing table sizes. Sometimes that 'x' would end up over the text.

johnsonm325:
In the Add System Modal, we get the following when you select the Baselines tab:

image

The filter bar and tabs are touching.
The modal size is not consistent between the Systems and Baselines tabs

johnsonm325:
I ran a comparison with a baseline, and the logo appears to be incorrect
image

johnsonm325:
Pending @katierik response to placement of the 'x', here's how it looks if only one system has been compared.
image

Doesn't look as good as it does when there are multiple systems, but I don't think people will be comparing 1 system. So it may be ok.

johnsonm325:
Lastly, are we supposed to have compact rows in the comparison table?

src/App.scss
```diff
@@ -76,41 +62,41 @@
...
-    position: relative;
-    height: 170px;
   }
 
   .add-system-header {
     background-color: #F5F5F5;
     padding: 20px 20px;
     text-align: center;
-    height: 170px;
+    //height: 50px;
```
johnsonm325: You pulled a Michael on this one.

ehelms:
Pushed a round of updates for nearly all things mentioned. I think I am still unclear on:

 * color of a systems timestamp when its "old"
 * if the placement of the 'x' is OK due to technology limitations

ehelms:
This can be re-reviewed now @johnsonm325

katierik:
@ehelms Placement of x is ok if we're stuck with it. 

Re: Single system - Lets set a max width for these systems. 500px should do it.

Re: color of timestamp when system is "old" - do you mean the timestamp on stale systems? The styling there is meant to exactly match the stale system styling on system lists.

ehelms:
I played around with the CSS and setting a max width doesn't seem to have any effect within a table.

katierik:
> I played around with the CSS and setting a max width doesn't seem to have any effect within a table.

Womp womp. The big stretch is alright then.

johnsonm325:
@ehelms 
In create baseline modal, the text input box spans the entire width of the modal. I don't think it looks good, but let me know what you think.
image

We also have a strange outline on the table for baseline facts:
image

And lastly, it looks like this visual update has adjusted the toolbar on the baselines page. The entire toolbar has shifted down and right, and the export button is butting up against the create baseline button:
image

ehelms:
Thanks for finding those issues, I've updated

johnsonm325:
@ehelms 
In this image, we don't have the yellow/gold color for the baseline, we don't have the correct icon for the baseline and we don't have a line separating the two systems in the header.
image

ehelms:
Good catch, fixed

PR #248, opened 2020-01-23
Change single select to radio buttons

johnsonm325:


johnsonm325:
I'll need to add some tests.

src/SmartComponents/BaselinesTable/redux/actions.js
```diff
@@ -15,14 +15,10 @@ function selectBaseline(ids, isSelected) {
...
-    return {
-        type: types.SELECT_ONE_BASELINE,
-        payload: { id, isSelected }
-    };
-}
-
 function setSelectedBaselines(selectedBaselineIds) {
+    /*eslint-disable*/
+    console.log('here');
+    /*eslint-enable*/
```
ehelms: there

ehelms:
Is it juts more or does the sorting on the baselines table inside the modal not work?

src/SmartComponents/BaselinesPage/BaselinesPage.js
```diff
@@ -38,23 +43,34 @@ export class BaselinesPage extends Component {
...
 
+    toggleModal = () => {
+        const { modalOpened } = this.state;
+        //const { fetchBaselines } = this.props;
+
+        this.setState({ modalOpened: !modalOpened });
+        if (modalOpened) {
+            //fetchBaselines();
+        }
+    }
```
ehelms: This function looks like it has a few leftovers from development.

src/SmartComponents/BaselinesPage/CreateBaselineModal/CreateBaselineModal.js
```diff
@@ -28,21 +29,34 @@ class CreateBaselineModal extends Component {
...
-                this.setState({ fromScratchChecked: false, copyBaselineChecked: false, copySystemChecked: true });
+                copySystemChecked = true;
             }
+
+            this.setState({
+                fromScratchChecked,
+                copyBaselineChecked,
+                copySystemChecked,
+                selectedBaseline: ''
+            });
```
ehelms: I believe you could simplify this functions logic with something like:

```
this.setState({
    fromScratchChecked: value === 'fromScratchChecked',
    ....
});
```

src/SmartComponents/BaselinesPage/CreateBaselineModal/CreateBaselineModal.js
```diff
@@ -28,21 +29,34 @@ class CreateBaselineModal extends Component {
...
         };
     }
 
+    isBaselineSelected(baselineId) {
+        this.setState({ selectedBaseline: baselineId });
```
ehelms: The action of this function doesn't fit with my expectation of it's name. The `is` portion of the function name makes me think this will return whether a baseline is selected or not rather than selecting a baseline.

src/SmartComponents/BaselinesPage/CreateBaselineModal/CreateBaselineModal.js
```diff
@@ -175,12 +187,12 @@ class CreateBaselineModal extends Component {
...
     renderActions() {
-        const { selectedBaselineIds, entities } = this.props;
-        const { baselineName, copyBaselineChecked, copySystemChecked } = this.state;
+        const { entities } = this.props;
+        const { baselineName, copyBaselineChecked, copySystemChecked, selectedBaseline } = this.state;
         let actions;
 
         if (baselineName === ''
-            || (copyBaselineChecked && selectedBaselineIds.length === 0)
+            || (copyBaselineChecked && selectedBaseline.length === 0)
```
ehelms: Is `selectedBaseline` now a baseline object? If so, I think this check should get updated to see if its undefined rather than checking its length.

src/SmartComponents/BaselinesTable/BaselinesTable.js
```diff
@@ -172,35 +157,74 @@ class BaselinesTable extends Component {
...
+                    isChecked={ data.selected }
+                    onChange={ this.onSingleSelect }
+                    name={ data[1] }
+                    id={ data[0] }
+                />
+            </React.Fragment>
+        );
+    }
+
+    createModalRows(modalData) {
```
ehelms: The table component shouldn't really need to know whether it's in a modal or not. So my assumption is that this is capturing a concept related to the Modal's use case but is bleeding the 'modal' terminology into the component. What is the functionality of the table itself that this is associated to? (e.g. single select)

src/SmartComponents/BaselinesTable/BaselinesTable.js
```diff
@@ -172,35 +157,74 @@ class BaselinesTable extends Component {
...
 
     renderTable() {
         const { baselineTableData, baselineListLoading, baselineDeleteLoading,
-            createBaselineModalOpened, addSystemModalOpened, hasSelect } = this.props;
+            isModal, addSystemModalOpened, hasSelect, modalTableData } = this.props;
         let columns = [
             { title: 'Name', transforms: [ sortable ]},
             { title: 'Last updated', transforms: [ sortable ]}
         ];
+        let radioColumns = [
```
ehelms: What is the difference between radio columns and other potential column sets?

src/SmartComponents/BaselinesTable/redux/actions.js
```diff
@@ -1,10 +1,10 @@
 import types from './types';
 import api from '../../../api';
 
-function fetchBaselines(params = {}) {
+function fetchBaselines(params = {}, isModal) {
```
ehelms: I think the idea of `isModal` is bleeding concepts into the layers that shouldn't care if it's a modal or not.

PR #251, opened 2020-01-30
Update comparison table to accept historical profiles

johnsonm325:


beav:
I was able to use the historical profile feature again with this patch. ack!

PR #252, opened 2020-02-04
Fix bugs with historical system profiles

johnsonm325:
Fixes remove button on historical profiles and correctly selects and deselects historical profiles in dropdown.


src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -218,9 +224,9 @@ class DriftTable extends Component {
             type = data[i].type;
 
             row.push(
-                <th key={ data[i].id } className={ type === 'baseline' ? 'drift-header baseline-header' : 'drift-header system-header' }>
+                <th key={ data[i].id } className={ data[i].type + '-header drift-header' }>
```
ehelms: You can do this with string interpolation as well:

```
className={ `drift-header ${data[i].type}-header` }
```

PR #254, opened 2020-02-05
Re-factor removeSystem function

ehelms:


ehelms:
Also fixes the icon for historical profile headers

johnsonm325:
Not sure if you've updated your PR on top of mine, but looks like a simple `npm run test -- -u` will work here to fix tests.

PR #264, opened 2020-02-14
Return empty array for systems, baselines and hspIds

johnsonm325:


beav:
This works for me but I get some console errors. I don't know if they are related to this PR or not.

```
Warning: Failed prop type: Invalid prop `onSelect` of type `boolean` supplied to `Table`, expected `function`.
```

```
Warning: Failed prop type: Invalid prop `rows[0]` supplied to `Table`.
```

```
Warning: Each child in a list should have a unique "key" prop. See https://fb.me/react-warning-keys for more information.
```

ehelms:
I think those ocnsole errors are normal right now. Looks like legit test failures though @johnsonm325

PR #266, opened 2020-02-17
Add error alert to baselines table

johnsonm325:


ehelms:
I merged the test PR which forced this to need a rebase.

src/SmartComponents/BaselinesPage/BaselinesPage.js
```diff
@@ -20,6 +23,38 @@ export class BaselinesPage extends Component {
         await window.insights.chrome.auth.getUser();
     }
 
+    renderCardBody = () => {
+        const { emptyState, baselineListLoading, fullBaselineListData } = this.props;
+
+        if (emptyState && !baselineListLoading) {
+            return this.renderEmptyState();
+        } else if (!baselineListLoading && fullBaselineListData.length === 0) {
```
ehelms: Do I understand correctly that this new display error is triggered if an API call is successful (a 200) but for some reason the data returned is empty? Under what conditions do we expect a successful API call that is empty yet is an error?

johnsonm325: `else if (!baselineListLoading && fullBaselineListData.length === 0)`
This basically means that the fetch is done, but nothing was returned. If nothing was returned, then it's a non-200.

ehelms: Should we be catching that from an HTTP response code perspective? I'm thinking if the API/reducer layer gets a response and that is not a 200, then we should be setting some variable on the state that indicates there was an error.

I worry that an empty data structure is not always indicative of an error happening. For example, will 300, 400 and 500s all return an empty list or will they return a body that has no data whatsoever? I think its good to differentiate those errors.

ehelms:
Will need a rebase

PR #268, opened 2020-02-19
Fixes data and header alignment with profiles

ehelms:
This includes a bit of re-factor towards having a single source of
truth for the master sorted system data.

### DIFF
--- a/src/SmartComponents/DriftPage/DriftTable/DriftTable.js
+++ b/src/SmartComponents/DriftPage/DriftTable/DriftTable.js
@@ -20,9 +20,12 @@ import { compareActions } from '../../modules';
 import { baselinesTableActions } from '../../BaselinesTable/redux';
 import { historicProfilesActions } from '../../HistoricalProfilesDropdown/redux';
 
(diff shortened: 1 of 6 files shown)


ehelms:
Updated, with a slight bump to coverage

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -385,16 +380,20 @@ class DriftTable extends Component {
     render() {
         const { fullCompareData, filteredCompareData, systems, baselines, historicalProfiles, loading } = this.props;
 
+        this.masterList = this.formatEntities(
```
johnsonm325: Because these are already loaded as constants, you could just `this.masterList = this.formatEntities(systems, baselines, historicalProfiles);`

ehelms: Good catch, this led me to notice that I missed updating two method signatures as well! So I simplified those -- the call to renderTable and renderRows

PR #271, opened 2020-02-21
Refactor BaselinesTable for radio buttons

johnsonm325:
Need some tests, but the refactor is good to test out.


src/SmartComponents/BaselinesPage/BaselinesPage.js
```diff
@@ -39,10 +80,24 @@ export class BaselinesPage extends Component {
...
+
         return (
             <CardBody>
                 <div>
-                    <BaselinesTable kebab={ true } createButton={ true } exportButton={ true } hasSelect={ true }/>
+                    <BaselinesTable
+                        kebab={ true }
+                        createButton={ true }
+                        exportButton={ true }
+                        hasSelect={ true }
```
ehelms: For me, `hasSelect` can be a bit confusing given multiselect and single-select are both forms of selection.

src/SmartComponents/AddSystemModal/AddSystemModal.js
```diff
@@ -92,7 +117,13 @@ class AddSystemModal extends Component {
                             eventKey={ 1 }
                             title="Baselines"
                         >
-                            <BaselinesTable hasSelect={ true }/>
+                            <BaselinesTable
+                                hasMultiSelect={ true }
```
ehelms: Is the default for BaselinesTable multi-select, single-select or no select?

src/SmartComponents/AddSystemModal/AddSystemModal.js
```diff
@@ -92,7 +117,13 @@ class AddSystemModal extends Component {
...
+                                tableType='CHECKBOX'
```
ehelms: What do you think about `tableId` rather than `type` since `type` tends to imply something that is from a defined set?

src/SmartComponents/AddSystemModal/AddSystemModal.js
johnsonm325: The default it whatever is passed to it. I actually needed to have removed the `hasMultiSelect` here. It makes it's decision based on the `tableType` passed to it.

src/SmartComponents/AddSystemModal/AddSystemModal.js
johnsonm325: so like: `tableId='CHECKBOX'`?

src/SmartComponents/AddSystemModal/AddSystemModal.js
```diff
@@ -92,7 +117,13 @@ class AddSystemModal extends Component {
...
+                                onSelect={ this.onSelect }
+                                baselineTableData={ baselineTableData }
+                                baselineListLoading={ baselineListLoading }
```
ehelms: I noticed that within BaselinesTable, `baselineListLoading` and `baselineDeleteLoading` are used in the same places and in the same way. This implies that they could be reduced down to a more generic `loading` variable and the interface here updated for that.

src/SmartComponents/BaselinesTable/BaselinesTable.js
```diff
@@ -135,30 +85,52 @@ class BaselinesTable extends Component {
...
-            let link = <div>
-                <a className="pointer active-blue"
-                    onClick={ () => this.fetchBaseline(baseline[0]) }
-                >
-                    { baseline[1] }
-                </a>
-            </div>;
-            row.push(link);
+
+            if (tableType === 'RADIO') {
```
ehelms: Should this be based on the type of select that's been configured for the component? `hasSingleSelect` ?

src/SmartComponents/AddSystemModal/AddSystemModal.js
```diff
@@ -92,7 +117,13 @@ class AddSystemModal extends Component {
                             eventKey={ 1 }
                             title="Baselines"
                         >
-                            <BaselinesTable hasSelect={ true }/>
+                            <BaselinesTable
+                                hasMultiSelect={ true }
+                                tableType='CHECKBOX'
+                                onSelect={ this.onSelect }
+                                baselineTableData={ baselineTableData }
```
ehelms: This could now be reduced to `tableData`

src/SmartComponents/BaselinesTable/BaselinesTable.js
```diff
@@ -312,36 +234,31 @@ BaselinesTable.propTypes = {
...
 };
 
 function mapStateToProps(state) {
     return {
         baselineDeleteLoading: state.baselinesTableState.baselineDeleteLoading,
-        baselineListLoading: state.baselinesTableState.baselineListLoading,
-        baselineTableData: state.baselinesTableState.baselineTableData,
-        addSystemModalOpened: state.addSystemModalState.addSystemModalOpened,
-        createBaselineModalOpened: state.createBaselineModalState.createBaselineModalOpened
+        addSystemModalOpened: state.addSystemModalState.addSystemModalOpened
```
ehelms: Looks like you can drop this and drop all of `mapStateToProps` which is a good sign that this component got dumber!

src/SmartComponents/BaselinesTable/BaselinesTable.js
```diff
@@ -312,36 +234,31 @@ BaselinesTable.propTypes = {
...
 }
 
 function mapDispatchToProps(dispatch) {
     return {
-        selectBaseline: (id, isSelected) => dispatch(baselinesTableActions.selectBaseline(id, isSelected)),
-        selectOneBaseline: (id, isSelected) => dispatch(baselinesTableActions.selectOneBaseline(id, isSelected)),
-        fetchBaselines: (params) => dispatch(baselinesTableActions.fetchBaselines(params)),
-        clearSelectedBaselines: () => dispatch(baselinesTableActions.clearSelectedBaselines())
+        clearSelectedBaselines: () => dispatch(baselinesTableActions.clearSelectedBaselines()),
+        fetchBaselines: (baselineType, params) => dispatch(baselinesTableActions.fetchBaselines(baselineType, params))
```
ehelms: Can both of these also be dropped thus dropping `mapDispatchToProps` ?

src/SmartComponents/BaselinesTable/redux/helpers.js
```diff
@@ -1,5 +1,22 @@
 import moment from 'moment';
 
+function fetchBaselines (tableType, fetchBaselines, fetchParams = {}) {
+    let params = {};
+
+    if (Object.entries(fetchParams).length !== 0) {
+        /*eslint-disable camelcase*/
+        params.order_by = fetchParams.orderBy;
```
ehelms: `.orderBy` could be undefined on the fetchParams object here. Same is true for any of the properties.

src/SmartComponents/BaselinesTable/redux/helpers.js
```diff
@@ -1,5 +1,22 @@
...
+        params.order_how = fetchParams.orderHow;
+
+        if (fetchParams.search) {
+            params.display_name = fetchParams.search;
+        }
+        /*eslint-enable camelcase*/
+    }
+
+    fetchBaselines(tableType, params);
+}
```
ehelms: Be cool to see a test for this nice generic function :)

src/SmartComponents/BaselinesTable/BaselinesTable.js
```diff
@@ -312,36 +234,31 @@ BaselinesTable.propTypes = {
...
     kebab: PropTypes.bool,
     createButton: PropTypes.bool,
     exportButton: PropTypes.bool,
     createBaselineModalOpened: PropTypes.bool,
-    hasSelect: PropTypes.bool,
-    hasSingleSelect: PropTypes.bool,
-    clearSelectedBaselines: PropTypes.func
+    onSelect: PropTypes.func,
+    clearSelectedBaselines: PropTypes.func,
+    columns: PropTypes.array
```
ehelms: Looks like some of the props in this object are now obsolete and could be removed.

src/SmartComponents/BaselinesTable/BaselinesTable.js
johnsonm325: That depends. I pass fetchBaselines to the helper function since I'm not sure how to dispatch an action from a helper without using mapDispatchToProps. Alternatively, I could drop it from BaselinesTable and pass the function in from 3 different parent Components.

src/SmartComponents/AddSystemModal/AddSystemModal.js
johnsonm325: As in, rename `baselineTableData` to `tableData`?

ehelms: Yea.

src/SmartComponents/BaselinesTable/BaselinesTable.js
ehelms: Ahh right, fair

ehelms:
Code looks good, the search filter isn't working on either table at the moment.

johnsonm325:
@ehelms should be working now

PR #273, opened 2020-02-25
Set timestamp to the same location within system header

ehelms:


johnsonm325:
@ehelms I want to make sure I'm seeing this correctly. Were you intending to create this much space between the system name and the timestamp?
image

ehelms:
Good catch! Updated to handle this more cleanly now.

PR #277, opened 2020-03-04
Allow only HSPs in comparisons

ehelms:


src/SmartComponents/DriftPage/DriftPage.js
```diff
@@ -123,7 +124,8 @@ function mapStateToProps(state) {
         error: state.compareState.error,
         loading: state.compareState.loading,
         systems: state.compareState.systems,
-        baselines: state.compareState.baselines
+        baselines: state.compareState.baselines,
```
johnsonm325: You can remove `systems` and `baselines` from here and `PropTypes`.

PR #279, opened 2020-03-05
Highlight rows that do not meet expectations red and highlight hover row

ehelms:


ehelms:
@johnsonm325 correct, will do a follow on once we have references in place to highlight individual cells

PR #286, opened 2020-03-27
Add ability to select system as a reference

johnsonm325:


johnsonm325:
Just remembered we'll have to wait for https://github.com/RedHatInsights/drift-backend/pull/152

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -91,7 +110,11 @@ export class DriftTable extends Component {
         }
         /*eslint-enable camelcase*/
 
-        return baselines.concat(fullHistoricalSystemList);
+        masterList =
+            this.referenceId
+                ? this.moveSystem(baselines.concat(fullHistoricalSystemList))
+                : baselines.concat(fullHistoricalSystemList);
```
ehelms: This performs a concat in both cases so I would opt to simplify this and move that up a line something like:

```
masterList = baselines.concat(fullHistoricalSystemLIst);

if (this.referenceId) {
  masterList = this.moveSystem(masterList);
}
```

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -38,13 +40,31 @@ export class DriftTable extends Component {
         await window.insights.chrome.auth.getUser();
         const { fetchCompare } = this.props;
 
-        fetchCompare(this.systemIds, this.baselineIds, this.HSPIds);
+        fetchCompare(this.systemIds, this.baselineIds, this.HSPIds, this.referenceId);
+    }
+
+    moveSystem = (systems) => {
```
ehelms: We are generally running into this problem, where we are dealing with baselines, systems and historical profiles. So the use of system nomenclature can get confusing. Naming ideas then `shiftReferenceToFront` and pass in `items` or `masterList`

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -38,13 +40,31 @@ export class DriftTable extends Component {
...
+
+        systems.forEach(function(system) {
+            if (referenceId === system.id) {
+                index = systems.indexOf(system);
+                systemToMove = system;
+            }
+        });
+
+        systems.splice(index, 1);
+        systems.unshift(systemToMove);
```
ehelms: I think you can simplify all of this logic a bit using some other builtins:

```
index = systems.findIndex(function (item) {
    return referenceId === system.id;
});

systemToMove = systems.splice(index, 1);
systems.unshift(systemToMove);
```

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -183,11 +216,23 @@ export class DriftTable extends Component {
...
+                        system.meets_reference_expectations || system.meets_reference_expectations === undefined
+                            ? 'comparison-cell'
+                            : 'highlight comparison-cell different-fact-cell' }
+                    >
+                        { system.value === null ? 'No Data' : system.value }
+                    </td>
+                );
+            } else {
+                row.push(
+                    <td className={ fact.state === 'DIFFERENT' ? 'highlight comparison-cell' : 'comparison-cell' }>
```
ehelms: Looks like the only difference between the two `row.push` statements is how this className is set. I think then you can extract that to logic thats wrapped in an if-else and call row.push only once.

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -256,6 +301,8 @@ export class DriftTable extends Component {
     renderSystemHeaders() {
         let row = [];
         let typeIcon = '';
+        let referenceId = this.referenceId;
+        let updateReferenceId = this.updateReferenceId;
```
ehelms: I sometimes find casting instance variables to local variables unnecessary. The additional use of `this.` in this function to reference those instance variables is easier to parse when reading the code.

src/SmartComponents/DriftPage/DriftTable/ReferenceSelector/ReferenceSelector.js
```diff
@@ -0,0 +1,38 @@
...
+    render() {
+        const { updateReferenceId, id } = this.props;
+        const { isReference } = this.state;
+
+        return (
+            isReference
+                ? <StarIcon
+                    className='reference-selector pointer'
+                    onClick={ () => updateReferenceId(undefined) }
+                />
```
ehelms: Should this be indented one more step to align with the `<StarIcon` ?

src/SmartComponents/DriftPage/DriftTable/ReferenceSelector/ReferenceSelector.js
```diff
@@ -0,0 +1,38 @@
...
+                : <OutlinedStarIcon
+                    className='reference-selector pointer'
+                    onClick={ () => updateReferenceId(id) }
+                />
```
ehelms: Should this be indented one more step to align with the `<StarIcon` ?

src/SmartComponents/DriftPage/DriftTable/ReferenceSelector/ReferenceSelector.js
```diff
@@ -0,0 +1,38 @@
...
+
+    render() {
+        const { updateReferenceId, id } = this.props;
+        const { isReference } = this.state;
+
+        return (
+            isReference
+                ? <StarIcon
+                    className='reference-selector pointer'
+                    onClick={ () => updateReferenceId(undefined) }
```
ehelms: You shouldn't need to call undefined explicitly here.

src/SmartComponents/DriftPage/DriftTable/ReferenceSelector/ReferenceSelector.js
johnsonm325: No, the `<StarIcon` isn't indented, it's just a space between the `?` and `<StarIcon`

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -177,17 +210,28 @@ export class DriftTable extends Component {
...
+                } else {
+                    className = 'comparison-cell';
+                }
+            } else {
+                if (fact.state === 'DIFFERENT') {
+                    className = 'highlight comparison-cell';
+                } else {
+                    className = 'comparison-cell';
+                }
+            }
```
ehelms, line 232: Note that the code adds `comparison-cell` in all cases. The `else` clauses could be dropped by starting with that:

```
let className = ['comparison-cell']

if (fact.state === 'DIFFERENT') {
     className.push('highlight');

     if (this.referenceId) {
         className.push('different-fact-cell');
}

row.push(<td className={ className,join() }>{ system.value === null ? 'No Data' : system.value }</td>);
```

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -38,13 +40,29 @@ export class DriftTable extends Component {
         await window.insights.chrome.auth.getUser();
         const { fetchCompare } = this.props;
 
-        fetchCompare(this.systemIds, this.baselineIds, this.HSPIds);
+        fetchCompare(this.systemIds, this.baselineIds, this.HSPIds, this.referenceId);
+    }
+
+    shiftReferenceToFront = (masterList) => {
+        let referenceId = this.referenceId;
```
ehelms, line 47: For me, this cast is not necessary and `this.referenceId` can be used in the body of the function.

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
johnsonm325: Great change!

johnsonm325:
Your logic change for `className` above didn't work like I thought it would, so I had to revert to something similar to what I already had.

PR #290, opened 2020-03-31
Fix: Update title from 'System Comparison' to 'Comparison'

CodeHeeler:


ehelms:
The build is green, not sure why its not reporting status https://travis-ci.org/github/RedHatInsights/drift-frontend/builds/669324226?utm_source=github_status&utm_medium=notification

PR #298, opened 2020-04-09
Disable Submit button on add system modal when nothing is selected

johnsonm325:


src/SmartComponents/AddSystemModal/AddSystemModal.js
```diff
@@ -96,7 +97,15 @@ export class AddSystemModal extends Component {
...
                             key="confirm"
                             variant="primary"
-                            onClick={ this.confirmModal }>
+                            onClick={ this.confirmModal }
+                            isDisabled={
+                                (entities && entities.selectedSystemIds && entities.selectedSystemIds.length === 0) &&
+                                selectedBaselineIds.length === 0 &&
+                                selectedHSPIds.length === 0
+                                    ? true
+                                    : false
```
ehelms, line 106: Given the condition of the ternary operator is a boolean, you can drop the ternary all together and simply pass the result of the expression to `isDisabled`.

PR #312, opened 2020-05-07
Rlong fix reducers

ryelo:


ryelo:
opened the PR in the wrong place oops

PR #330, opened 2020-07-18
Update Baseline pagination to ensure using from the backend

ehelms:


ehelms:
Add 'Select none (0)' and 'Select page (50)'

ehelms:
Moving between pages is broken

PR #331, opened 2020-07-18
Display tag filtering in systems table

ehelms:
@karelhala Could you verify this is the correct way to display tag filtering in our use of the inventory table?


karelhala:
Yup, that should do the trick.

PR #338, opened 2020-08-07
Add release-qa and release-prod scripts

ehelms:


scripts/release-prod
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -e
+
+read -p "Releasing master(CI) to QA environment. Do you wish to continue? (Y/n) " decision
```
johnsonm325, line 3: Shouldn't this be "Releasing qa to prod environment"? Or something like it?

ehelms: Good catch. Updated.

PR #340, opened 2020-08-13
[Beta] change same/different filters to expected/unexpected

CodeHeeler:


src/SmartComponents/modules/reducers.js
```diff
@@ -11,11 +11,19 @@ const initialState = {
...
+            { filter: 'DIFFERENT', display: 'Unexpected', selected: true },
+            { filter: 'INCOMPLETE_DATA', display: 'Uncertain', selected: true }
+        ]
+    } else {
+        stateFilters: [
+            { filter: 'SAME', display: 'Same', selected: true },
+            { filter: 'DIFFERENT', display: 'Different', selected: true },
+            { filter: 'INCOMPLETE_DATA', display: 'Incomplete data', selected: true }
+        ]
+    },
```
ehelms, line 26: This if won't work inside the hash declaration. You'll want to move it after wards. So having the hash have sane defaults (the way it was prior to the change) and then modifying, e.g.

```
const initialState = {
    fullCompareData: [],
    filteredCompareData: [],
    sortedFilteredFacts: [],
    referenceId: undefined,
    systems: [],
    baselines: [],
    historicalProfiles: [],
    previousStateSystems: [],
    stateFilters: [
            { filter: 'SAME', display: 'Same', selected: true },
            { filter: 'DIFFERENT', display: 'Different', selected: true },
            { filter: 'INCOMPLETE_DATA', display: 'Incomplete data', selected: true }
    ],
    factFilter: '',
    totalFacts: 0,
    page: 1,
    factSort: ASC,
    stateSort: DESC,
    perPage: 50,
    loading: false,
    expandedRows: [],
    error: {},
    emptyState: true
};

if (insights.chrome.isBeta()) {
        initialState['stateFilters'] = [
            { filter: 'SAME', display: 'Expected', selected: true },
            { filter: 'DIFFERENT', display: 'Unexpected', selected: true },
            { filter: 'INCOMPLETE_DATA', display: 'Uncertain', selected: true }
        ]
};

```

src/SmartComponents/modules/reducers.js
```diff
@@ -28,6 +28,14 @@ const initialState = {
     emptyState: true
 };
 
+if (insights.chrome.isBeta()) {
+    stateFilters: [
```
ehelms, line 32: `initialState['stateFilters'] = [`

CodeHeeler:
So it works locally as expected.  For reference, to get past the errors and make eslint happy, I had eslint --fix take care of the file and it ends up with a weird thing you'll see in reducer.js where there's a new line between a } and a ; with the semicolon just dangling there by itself which I really do not like.  Googling around, it looks like there's a lot of arguments on the way eslint does things so it's not just us, but it does work, I just wanted you aware of the weirdness and why.

johnsonm325:
Some failing tests. Easy to fix.

What you'll need to do is run `npm run test` in your terminal to see the tests that are failing. In each of those test files, you'll need to create `insights.chrome.isBeta()`. Those tests are failing because in this mocked instance, there is no such thing as insights, so it can't find it, much less chrome or isBeta.

Basically, insert this into each of the files after imports, but before the tests are described. Kind of like you see here: https://github.com/RedHatInsights/drift-frontend/blob/master/src/SmartComponents/BaselinesPage/__tests__/BaselinesPage.tests.js#L12

```
global.insights = {
    chrome: {
        isBeta: jest.fn()
    }
};
```

CodeHeeler:
@johnsonm325 Ok, I've added the global.insights snippet as you advised, thanks!  I'm still getting the tests failing locally on the same isBeta spots.  Is there something I need to do to ensure` npm run test` picks up the latest changes?  I've pushed up the changes so you can take a look as well.

johnsonm325:
Here's the fix! Update config/setupTests.js to look like this:
```
import { configure, mount, render, shallow } from 'enzyme';
import Adapter from 'enzyme-adapter-react-16';
import React from 'react';

configure({ adapter: new Adapter() });

global.shallow = shallow;
global.render = render;
global.mount = mount;
global.React = React;

global.insights = {
    chrome: {
        auth: {
            getUser: () => new Promise((resolve) => {
                setTimeout(resolve, 1);
            }),
            logout: jest.fn()
        },
        isBeta: jest.fn()
    }
};
```

Afterwards, we can remove the `global.insights` set ups in the tests across the app. You can do a search, but I'm only seeing it in these files:

- https://github.com/RedHatInsights/drift-frontend/blob/master/src/SmartComponents/DriftPage/__tests__/DriftPage.tests.js#L11
- https://github.com/RedHatInsights/drift-frontend/blob/master/src/SmartComponents/DriftPage/DriftFilterChips/__tests__/DriftFilterChips.tests.js
- https://github.com/RedHatInsights/drift-frontend/blob/master/src/SmartComponents/DriftPage/DriftTable/__tests__/DriftTable.tests.js#L15
- https://github.com/RedHatInsights/drift-frontend/blob/master/src/SmartComponents/modules/__tests__/reducer-test.js
- https://github.com/RedHatInsights/drift-frontend/blob/master/src/SmartComponents/AddSystemModal/__tests__/AddSystemModal.tests.js
- https://github.com/RedHatInsights/drift-frontend/blob/master/src/SmartComponents/BaselinesPage/__tests__/BaselinesPage.tests.js#L12
- https://github.com/RedHatInsights/drift-frontend/blob/master/src/SmartComponents/BaselinesPage/CreateBaselineModal/__tests__/CreateBaselineModal.tests.js
- https://github.com/RedHatInsights/drift-frontend/blob/master/src/SmartComponents/BaselinesPage/EditBaseline/__tests__/EditBaseline.tests.js#L13

PR #341, opened 2020-08-24
Refactor error alerting for entire application

johnsonm325:
Easiest way to test is updating this file (https://github.com/johnsonm325/drift-frontend/blob/33ce91e75cac2354b27215a2bea7b42cc786275c/src/api.js) like so:

```
import axios from 'axios';
import { DRIFT_API_ROOT, BASELINE_API_ROOT, HISTORICAL_PROFILES_API_ROOT } from './constants';

async function post(path, body = {}) {
    const request = await axios.post(DRIFT_API_ROOT.concat(/*path*/'blah'), body);
    return request.data;
}

async function getBaselines(path, getParams = {}) {
    const request = await axios.get(BASELINE_API_ROOT.concat(/*path*/'blah'), { params: getParams });
    return request.data;
}

/*eslint-disable no-unused-vars*/
async function getBaseline(path) {
/*eslint-enable no-unused-vars*/
    const request = await axios.get(BASELINE_API_ROOT.concat(/*path*/'blah'));
    return request.data.data[0];
}

async function patchBaseline(path, body = {}) {
    let request = await axios.patch(BASELINE_API_ROOT.concat(/*path*/'blah'), body);

    return request;
}

async function postBaseline(path, body = {}) {
    const request = await axios.post(BASELINE_API_ROOT.concat(/*path*/'blah'), body);
    return request.data;
}

async function deleteBaselines(path, body = {}) {
    let response = await axios.post(BASELINE_API_ROOT.concat(/*path*/'blah'), body);

    return response;
}

/*eslint-disable no-unused-vars*/
async function getHistoricalData(path) {
/*eslint-enable no-unused-vars*/
    let response;

    const request = await axios.get(HISTORICAL_PROFILES_API_ROOT.concat(/*path*/'blah'))
    .catch(function (error) {
        return error.response;
    });

    if (request.status === 200) {
        response = request.data.data[0];
    } else {
        response = request;
    }

    return response;
}
```

This will force errors for the api requests. You'll have to revert some of the code back to get to the next step (ie you'll need to fix `getBaselines` in order to test `getBaseline`)

At the most basic level, we need to test that errors on fetching result in an empty state with a toast notification. Errors in modals result in inline errors on the modal. And errors on deleting show a toast notification. If you have specific questions we can chat.


src/SmartComponents/BaselinesPage/EditBaseline/EditBaseline.js
```diff
@@ -236,30 +284,50 @@ export class EditBaseline extends Component {
...
-                        text={ [ 'No facts or categories have been added to this baseline yet.' ] }
-                        button={ <AddFactButton /> }
-                    />
-                </td>;
         }
 
         return rows;
     }
 
+    renderEmptyState() {
```
bastilian, line 292: Have you thought of extracting this into a separate component?
It looks very similar to what is on https://github.com/RedHatInsights/drift-frontend/pull/341/files#diff-ff14a549333021d3958a55730a76b8dbR107 and maybe could be used in both places.

src/SmartComponents/BaselinesPage/EditBaseline/EditBaselineNameModal/EditBaselineNameModal.js
```diff
@@ -101,6 +101,18 @@ export class EditBaselineNameModal extends Component {
...
+                        variant='danger'
+                        isInline
+                        title={ 'Status: ' + error.status }
+                    >
+                        <p>
+                            { error.detail }
+                        </p>
+                    </Alert>
+                    : <div></div>
+                }
```
bastilian, line 115: This could be written as just:

```js
{ error.status && <Alert
  variant='danger'
  isInline
  title={ 'Status: ' + error.status }>
    <p>
      { error.detail }
    </p>
</Alert> }
```

src/SmartComponents/ErrorAlert/ErrorAlert.js
```diff
@@ -1,90 +1,54 @@
...
+                { error.status
+                    ? addNotification({
+                        variant: 'danger',
+                        title: 'Error',
+                        description: error.detail,
+                        dismissable: true
+                    })
+                    : null
+                }
+                { error.status ? this.closeToast(tableId) : null }
```
bastilian, line 35: I'm a bit confused. :) Does this component render something or is it only to call functions when `error.status` is set?

johnsonm325: This component calls the `addNotification` in frontend-components/notifications whenever the error object that is passed in has a status code. I don't pass an error with a status of 200, so it would only ever get an error status.

src/SmartComponents/BaselinesPage/EditBaseline/EditBaseline.js
bastilian: Yeah, I meant creating another component to handle which EmptyStateDisplay to show. But looking at it closer it seems to just differ in the props it needs.

In compliance we try to avoid having functions in components that return components and prefer to extract them into a separate (functional) component. For readability and since we have seen components built that way act unexpected.

bastilian:
I tried testing this by creating a Baseline and change the `api.js` to request a non-existent path for the delete function before clicking the delete in the UI. It did throw an error (404) in the console, but it didn't show any Alert. 

Should it be there or is that the wrong place to look?

johnsonm325:
@bastilian ahh, I need to rename this PR. This started as a PR to show an error alert for deleting a baseline, but ended up being a refactor of error alerting. That being said, I did just test it and got the toast alert. Perhaps we should chat about it? I'll ping you.

src/SmartComponents/BaselinesPage/EditBaseline/EditBaselineNameModal/EditBaselineNameModal.js
```diff
@@ -62,14 +62,14 @@ export class EditBaselineNameModal extends Component {
                     isRequired
                     fieldId='baseline name'
                     helperTextInvalid={ error.hasOwnProperty('detail') ? error.detail : null }
-                    isValid={ !error.hasOwnProperty('status') }
+                    validated={ error.hasOwnProperty('status') ? 'error' : null }
```
bastilian, line 65: You could use `ValidatedOptions.error` here as well, and move this into a `const` to use here and on the `TextInput`.

bastilian (approved):
Works for me! I tested as @johnsonm325 suggested by modifying the `api.js` and notifications as well as inline errors appear as expected!

eherget (approved):
Approving for Sebastian, who doesn't have write access.

PR #342, opened 2020-08-27
Add header to HSP dropdown for systems

johnsonm325:
Also fixes spacing issues on dropdown scroll and padding all around dropdown


eherget:
Looks good and it works!  merging...

PR #346, opened 2020-09-09
Fixes Empty system names in create baseline modal

johnsonm325:
Steps to reproduce:
1. Baselines -> Create baseline -> Copy an existing system
2. Find a system with hsps and select an hsp, but don't create the baseline
3. Select Copy an existing baseline
4. Select Copy an existing system. The name column is blank for all systems.


CodeHeeler (approved):
Tested locally and it works! Thank you.

PR #347, opened 2020-09-11
Change export button to dropdown

johnsonm325:
Export to CSV should now be in a dropdown (where we will eventually add Export to JSON as well). The dropdown should be tested on Comparison, Baselines, and edit Baseline pages.


CodeHeeler (approved):
Works for me locally in Comparison, Baseline, and Create Baseline, both visually seeing the dropdown, and then clicking to Export to CSV still working and downloading correctly as well.
image

PR #348, opened 2020-09-14
Clear selected HSPs on 'Clear comparison'

johnsonm325:


CodeHeeler (approved):
Tested locally, works correctly unselecting hsps after the clear all comparisons button is clicked.
image

PR #349, opened 2020-09-17
First baseline in comparison gets automatic reference set

johnsonm325:
Build a comparison with at least one baseline:
- The first baseline in the comparison gets set as the reference (look for the star to be highlighted on the baseline header)
- If you choose something else as a reference, it will de-reference that baseline. De-referencing your second choice of a reference will result in no references being set

Build a comparison with anything but a baseline:
- If you build your comparison without the baseline, and then add a baseline to the comparison, the baseline will be set at the default reference.
- If you build your comparison without the baseline, then choose one of your systems/HSPs as a reference, then dereference it, and then add a baseline, that baseline will not be set as the default reference

### DIFF
--- a/src/SmartComponents/DriftPage/DriftTable/DriftTable.js
+++ b/src/SmartComponents/DriftPage/DriftTable/DriftTable.js
@@ -28,7 +28,8 @@ export class DriftTable extends Component {
             emptyStateMessage: [
                 'You currently have no system or baselines displayed. Add at least two',
                 'systems or baselines to compare their facts.'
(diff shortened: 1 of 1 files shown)


CodeHeeler:
The first and third cases above work as expected.  In the second case, choosing another reference does properly de-reference the first baseline, however, after de-referencing that second choice, instead of no reference, it goes back to having that initial baseline set as reference.

CodeHeeler (approved):
Everything works as expected now, great job!

PR #350, opened 2020-09-29
Add rbac permissions to frontend

johnsonm325:
Some tests are skipped due to a lack of support for shallow rendering in Enzyme for context provider and consumers. Couldn't get some of the tests to pass, so I'll work on fixing them at another time.

To test, go to `src/App.js` and change the values of permissions to false as needed: https://github.com/RedHatInsights/drift-frontend/blob/3731bc7115cc5ed362403090fb3e04df5d4ffa3b/src/App.js#L68-L70

To test compare read permissions:
- Set compareRead to false
- Go to drift comparison page and test that you get an empty state, keeping you from performing a comparison

To test baselines read permissions:
- Set baselinesRead to false
- Go to baselines page and test that you get an empty state, keeping you from viewing a list of baselines
- Go to a page to view a single baseline and test that you get an empty state, keeping you from viewing the baseline facts. You should be able to go back to the baselines page via the Breadcrumb
- You should not be able to edit the name of the baseline

To test baselines write permissions:
- Set baselinesWrite to false
- Go to baselines page and test that the selection checkboxes are gone, kebabs for individual baselines are gone, bulk select is disabled and create baseline button is disabled with tooltip
- Go to a page to view a single baseline and test that the selection checkboxes are gone, kebabs for individual facts are gone (expanded categories and check), bulk select is disabled and Add fact button is disabled with tooltip
- Go to drift comparison page and try to add a baseline. You should be able to select baselines to add.

Feel free to test outside of these bounds as well.


CodeHeeler (approved):
Tested locally and it functions as expected, great work thank you!

PR #351, opened 2020-09-30
Fix text input error alert in fact modal

johnsonm325:
To test, try to create a new fact with the same name as an existing fact.


CodeHeeler (approved):
Looks great, before there was no error, just nothing happened, now it is very informative:
image

PR #352, opened 2020-09-30
Add html attribute to sorting widget to ease automation development

johnsonm325:


CodeHeeler (commented):
you have a id={ stateSort } to go with your id='', but I don't see an equivalent one for the new factSort, is that expected?

johnsonm325:
Added attribute to factSort and added tests

CodeHeeler (approved):
This looks great now AFAICT, and thank you for adding test coverage!

PR #356, opened 2020-10-05
First reference reset on clear all comparisons and removing all compa…

johnsonm325:
…risons


CodeHeeler (approved):
The functionality here now works as expected, resetting and adding again yields a referenced baseline again.

PR #357, opened 2020-10-06
Pass tags to inventory component

karelhala:


karelhala:
ping @johnsonm325

johnsonm325 (changes requested):
.

src/App.js
```diff
@@ -1,85 +1,79 @@
...
-                    <NotificationsPortal />
-                    <Routes childProps={ this.props } />
-                </PermissionContext.Provider>
-                : null
-        );
-    }
-}
+        insights.chrome.on('GLOBAL_FILTER_UPDATE', ({ data }) => {
+            const tags = insights.chrome?.mapGlobalFilter?.(data).filter(item => !item.includes('workloads')) || undefined;
+            dispatch(actions.setGlobalGilterTags(tags));
```
johnsonm325, line 58: Should this be `setGlobalFilterTags`?

src/SmartComponents/modules/actions.js
```diff
@@ -101,6 +101,13 @@ function updateReferenceId(id) {
     };
 }
 
+function setGlobalGilterTags(tags = []) {
```
johnsonm325, line 104: `setGlobalFilterTags`?

johnsonm325 (changes requested):
@karelhala I had to do some research on hooks to make sure I understood what was happening. I'll need help testing this. I'm not sure what has to happen to get `GLOBAL_FILTER_UPDATE` to fire off.

src/App.js
karelhala: Yup, good catch!

karelhala:
@johnsonm325 sure thing! This is to enable the global filter feature, basically allowing users to set (currently) tags that they want the UI to be prefiltered by for all applications. If you want to see the global filter on local development you have to fire `insights.chrome.enable.globalFilter()` in console and refresh page. Currently the global filter is allowed only on beta environments and this will set localStorage key to enable it ignoring beta environment. After that you can select any tag and it will prefilter the systems modal for comparision. Basically this ticket https://issues.redhat.com/browse/RHCLOUD-9580

johnsonm325 (approved):

PR #358, opened 2020-10-06
Fire proper clear all filters event on modal load

karelhala:
When systems modal is opened inventory event `CLEAR_ALL` is fired, this causes inventory to incorrectly load systems as it ignores `source=insights` if filters are cleared. This PR adds new event that clear all filters from drift perspective, meaning it will set `activeFilters` to undefined so inventory will respect default filters and shows correct systems. This event is fired only on modal render so if user clears filters in modal proper systems are shown and when some filter is applied and modal is closed and re-opened this modal shows again systems for default filter.


karelhala:
ping @johnsonm325

johnsonm325:
@karelhala You'll have to help me understand how to test this. I must be doing something wrong or missing the point because I'm not seeing a difference.

karelhala:
@johnsonm325 well if you open network tab in development console and open systems modal you can see that without this change fetching hosts has no `source` filter defined, with this patch it will respect this filter and prefilter the systems based on correct source that is also visible in filters.

johnsonm325:
Here's the request URL on master: `https://ci.foo.redhat.com:1337/api/inventory/v1/hosts?per_page=50&page=1&order_by=updated&order_how=DESC&staleness=fresh&staleness=stale&&registered_with=insights`
Here's the request URL on this PR: `https://ci.foo.redhat.com:1337/api/inventory/v1/hosts?per_page=50&page=1&order_by=updated&order_how=DESC&staleness=fresh&staleness=stale&&registered_with=insights`

Is this what you're referring to?

karelhala:
Running correct travis

johnsonm325 (approved):

sthirugn:
@karelhala when do you want this in prod?

PR #359, opened 2020-10-06
Fix no empty state when viewing baseline

johnsonm325:
To reproduce:
1. View a baseline with no facts (empty baseline)
2. Attempt to edit the baseline name, but simply cancel the edit
3. You shouldn't see the empty state for the baseline, but instead an empty table


CodeHeeler (approved):
Now the empty state shows correctly as opposed to the empty table after cancelling an edit to the baseline name.  Thanks!

PR #360, opened 2020-10-08
Fix copy existing system modal from disappearing

johnsonm325:
Steps to reproduce:
- Navigate to Baselines
- Click Create baseline
- Select Copy an existing baseline
- Enter baseline name
- Select Copy an existing system
- The modal disappears, an empty page is displayed..


CodeHeeler (approved):
The copy to create baselines now works correctly without the modal disappearing.

PR #361, opened 2020-10-08
Fixes baselines table sorting with drift:baselines:read only permissions

johnsonm325:
I'm only seeing this as a read-only user. To repro:

- Restrict user to read-only for baselines
- Click on Baselines (all are displayed)
- Click 'Name'. A 400 error is displayed


CodeHeeler (approved):
Sorting by Name works correctly now even with no write permission for baselines.

PR #362, opened 2020-10-09
Add fact button disabled no write rbac permissions/empty baseline

johnsonm325:
Set rbac permissions for baselines write to false. To repro:
1. Select or create an empty baseline
2. The 'Add fact or category' button will be enabled, when it should be disabled

This change should make the button disabled with a tootip.

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -364,3 +364,7 @@
   height: 18px;
   margin: 0px;
 }
+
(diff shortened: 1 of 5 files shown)


CodeHeeler (approved):
Yes, the button is disabled and has an added tooltip w/ no baselines write permission.

PR #365, opened 2020-10-14
Fix broken rbac permissions

johnsonm325:
Updated so rbac permission checking is only done in beta.


beav (approved):
ack from me. The change looks reasonable and its to fix a broken state

CodeHeeler (approved):
Thanks for the fix!

PR #366, opened 2020-10-14
Fix more issues with rbac

johnsonm325:
I wasn't passing props from the app level all the way down to Baselines Table through Add System Modal. Two different components build Baselines Table and I was only passing it down one of those paths.


CodeHeeler (approved):
Glad you figured it out, thank you so much!

PR #368, opened 2020-10-22
Fix disabled create baseline button if there are no existing baselines

johnsonm325:
When there is an empty baselines list and the user has baselines write permissions, the create baselines button is disabled when it should not be.


eherget (approved):
I have tested this fix and it works in my local environment.  As soon as the travis checks pass, go ahead and merge.

PR #369, opened 2020-10-22
Baseline name doesn't show with no read permissions

johnsonm325:
To reproduce:
1. Make sure you don't have baseline read permissions and try to view a baseline
2. Breadcrumb and title should show the name of the baseline

This fix will ensure that the breadcrumb only shows a link back to baselines list and the title will say 'Baseline'.


CodeHeeler (approved):
It works as expected now, with no name in the baseline nor breadcrumb.  Thanks!

PR #370, opened 2020-10-22
Fix bulk select in Systems Table

johnsonm325:
Bulk selector in systems table on add systems modal should work to be able to select an entire page of systems. If you select an entire page and then go to the next page, those selections will still be selected. It keeps a running total. Because the api requests paginated data, you cannot select the entire inventory system at once.

Definitely mess around with this and try to break it. Try adding filters, sorting and selecting/deselecting things to make sure I'm not missing a key part here.

The bulk selector in the systems table on the create baseline modal should be disabled as you can only select a single system to copy from.


eherget (approved):
Works for me and looks good. Let's see if QE can break it!

PR #371, opened 2020-10-23
Enable global filter SID and SAP flag

karelhala:
### Fully integrate drift to global filter

As we approach further the monday global filter release we have to consume SID and workloads values from global filter. Also start using new version of chrome global filter mapper.


karelhala:
ping @johnsonm325

karelhala:
Running correct travis

johnsonm325 (approved):

sthirugn:
@karelhala when do you want this in prod?

karelhala:
@sthirugn next monday we will be releasing all other apps but since drift is partially integrated we can release anytime you feel good about it.

johnsonm325:
@karelhala Needs rebase

PR #372, opened 2020-10-26
Baselines list loads on breadcrumb click

johnsonm325:
To reproduce:
- Go to Baselines page with an empty state (no baselines)
- Create a baseline
- From the edit baseline page, click the breadbrumb to go back to the baselines list

The baselines list will still show empty. If you refresh, you'll see the baseline you just created.

This fix will show the new baseline you created after you click the breadcrumb.


eherget (approved):
This works!

PR #373, opened 2020-10-27
Update HSP badge count on Bulk Select 'Select none'

johnsonm325:
- Open Add to comparison modal
- Select several HSPs
- Click Select all
- Click Deselect all

The HSP selector shows no. of selected HSPs even though none are selected.

This fix updates the badge count when using the bulk select 'Select none' to 0.


eherget (changes requested):
With this PR's code in place, clicking on the HSP checkbox causes the modal to disappear and.. wow, this is weird... I'll start a google meet so you can see.

eherget (approved):

PR #374, opened 2020-10-27
Use stable version of PF

karelhala:
When using inventory filter tags it blows up because PF is trying to access to object partial which doesn't exists. This PR fixes such issue by increasing version to known stable version.


karelhala:
ping @johnsonm325

johnsonm325 (approved):

PR #376, opened 2020-10-28
Change state icons to 16x16 pixels

johnsonm325:


CodeHeeler (approved):
If this is expected, approved, but where your code says 16, the icon itself says 15.5, see attached image.  I wasn't sure which one needed to be 16 or if they usually don't match like this.
image

CodeHeeler:
Katie Riker said this looks good!

PR #377, opened 2020-10-28
Fix resizing of baseline table 'Last updated' column

johnsonm325:


CodeHeeler (changes requested):
I'm still seeing this jumping to the left behaviour with the Last Updated sort name.

CodeHeeler (approved):
fixed and/or the fault was mine: when full screen the behaviour is as expected, when my browser is my usual half screen, the jumping still happens, but I know that there are tons of things broken for that sized window and that we're not yet worrying about those cases.

PR #378, opened 2020-10-29
Fix issue with removing system as reference

johnsonm325:
Two issues were fixed here.
First:

- Add two systems to comparison
- Make one a reference
- Remove the system that is the reference
Get an error.

Second:

- Select two systems for comparison
- Make one of them the reference
- Remove the other, then remove the reference host
An error: "must specify at least one of system, baseline, or HSP" is displayed. See screenshot.


CodeHeeler (approved):
This fixes the issues, no errors occur for either case now.  Thanks!

PR #379, opened 2020-10-29
Remove hover state from edit baseline table

johnsonm325:


CodeHeeler (approved):
Hover state is now gone, thanks!

PR #380, opened 2020-10-29
Fix spacing on Fact Modal

johnsonm325:
Spacing between fact and value text input areas are decreased.


CodeHeeler (approved):
The spacing is correct now, thanks!

PR #381, opened 2020-10-29
Fix spacing in Edit Baseline Name modal under title

johnsonm325:
Also fixed spacing of form groups in edit baseline name modal and edit/add fact modals


CodeHeeler (approved):
I can't tell if this is different or not?  I think it looks great, so approved assuming anyone can tell from these screenshots:
image
image

PR #383, opened 2020-10-30
Relocate loading icon on create baseline modal

johnsonm325:
If you click 'copy from an existing system' on the create baseline modal, you'll see the loading icon beside the text 'copy from an existing system'. This PR moves that icon below the text.


CodeHeeler (approved):
This does correctly move the spinner beneath the text.  The card says "centered below" and this is not centered.  If that doesn't matter, approved, otherwise, update to center first.
image

katierik:
Ok with it being not centered!

PR #384, opened 2020-11-02
Update PF Icons to 4.7.16

johnsonm325:
Among some other changes, we should see a change in the blueprint icon. The blueprint icon can be seen in the header of a comparison next to a baseline name.

1. Create a comparison with a baseline
2. Click Submit
3. See the icon to the left of the baseline name

It should look different from the icon on ci.cloud. It should not be cut-off the same way. Refer to the Trello card for specifics and compare ci.foo to ci.cloud and zoom it. It should be relatively obvious, but let me know if you have questions.

Old (incorrect) icon:
image

New (correct) icon:
image


CodeHeeler (changes requested):
I'm still seeing the old incomplete icon:
image

CodeHeeler (approved):
works perfectly!  thanks for reminding me to rm node modules and rerun npm i.

PR #385, opened 2020-11-02
Create Drift Toolbar component

johnsonm325:
To test, just make sure that all of the functionality of the drift comparison toolbar works as expected.
- Filtering
- State filtering
- Add to comparison
- Export to CSV
- Clear all comparisons
- Pagination
- Filter chips
- Clear filters


CodeHeeler (approved):
all features still work as expected, thanks!

PR #386, opened 2020-11-03
Fix error on header of baseline table in add system modal

johnsonm325:
This fix should allow you to load inventory table. Make sure to check baseline table in add system modal.


eherget (approved):
Looks good and works for me!

PR #387, opened 2020-11-04
Remove break from fact modal

johnsonm325:


eherget (approved):
Looks good and works!

PR #388, opened 2020-11-05
Break comparison header into its own component

johnsonm325:
Header in comparison table should act the same as it did before.


CodeHeeler:
Here's the before and after, there is a missing vertical line between the text box labelled 'Search tags' and the down arrow.
Before:
image
After:
image

CodeHeeler (changes requested):
I don't see any changes to functionality, so I think you're good once you fix the one visual detail commented on.

CodeHeeler (approved):
The noted discrepancy is how it is expected to look across apps, and all functionality is the same, so approved!  Thanks for the explanation and the work.

PR #390, opened 2020-11-11
Refactor Baselines Table

johnsonm325:
Update some of the properties in BaselinesTable and use SkeletonTable from frontend-components in place of loading rows for multiple files.

For SkeletonTable, make sure the correct box show up for each row (checkbox vs radio) and make sure they are disabled during loading. Check these pages for the Skeleton table:
- BaselinesTable (Baselines page, create baseline modal and add system modal)
- EditBaseline

The loading table for baselines on the add system modal will look broken until I make another change on a different PR. I haven't started working on that one because I wanted to put this one through first.


CodeHeeler (approved):
Perfect!  Thank you!

PR #394, opened 2020-11-25
Refactor toolbar to fix chips padding

johnsonm325:


CodeHeeler:
I'm not seeing any difference in spacing, but I'll check in with you tomorrow and check again.

CodeHeeler (approved):
I see it now, it was subtle, thanks for the clarification of what to expect!

PR #395, opened 2020-11-25
Remove isBeta check on RBAC

johnsonm325:


petrblaho (approved):
Code looks ok (what can go wrong with just removing one `if`? :-))

I have no idea how I can test it though...

PR #396, opened 2020-11-25
Refactor chips on baselines table and fix padding

johnsonm325:


CodeHeeler:
For this one, I only see a new message about the number of filters but no chips like there were before:
image

johnsonm325:
@CodeHeeler You've got your page at a half-screen size. It does that when the screen is smaller. If you make it the full screen size you should see the filter.

CodeHeeler (approved):
LGTM (at full screen)

PR #398, opened 2020-11-30
Enable Add Fact Button when baseline only has an empty category

johnsonm325:
Steps to reproduce:
- Create a baseline from scratch
- Add a category to the baseline

The add fact or category button is inactive


petrblaho (approved):
LGTM!

PR #400, opened 2020-12-01
Add tooltips to icons in comparison table

johnsonm325:
Tooltips should be on the following:
- State icons
    - Tooltips should read differently for same, different and incomplete data
    - Different tooltips should be different if a reference is set (Same and incomplete remain unchanged)
    - Tooltips should reference category vs. row if it is a category
- Icons next to baseline/system/hsp name
- Reference selector
    - Tooltip should read differently on set reference vs. unset reference


petrblaho (approved):
It shows tooltips on icons exactly as mentioned in description.

Looks good to me!

petrblaho (approved):

PR #401, opened 2020-12-04
Add single select option to baseline table

johnsonm325:


petrblaho:
Is this only implementation changes / refactor? IE no visible or behavior changes?

petrblaho:
I was confused a lot by `tableId` variable.
It is used in components as a type of table(rows) but the name suggests it is an identifier of the table. As I understand we do not have more tables in the app right now so it should not be a big issue.

CodeHeeler (approved):
This works as it did before, I have not found any issues with it.  Thanks!

petrblaho (approved):

PR #402, opened 2020-12-07
Change 'Fact name' to 'Baseline name' on baseline table chip

johnsonm325:
If you filter the baselines by name, the chip group that appears in the toolbar will read "Fact name". With this fix, it should read "Baseline name".


petrblaho (approved):
LGTM!

PR #403, opened 2020-12-08
Fix break on 'Clear all filters' drift table

johnsonm325:
To reproduce:
1. Add a system to a comparison that contains a category with state SAME and at least 1 INCOMPLETE_DATA. The state on the category will be DIFFERENT in this case.
2. Click 'Clear all filters' link next to the chips.

This should cause an error. You may not see it, but the Incomplete data chip should remain in the chips, which is not what we expect. We expect to see all chips cleared out.

This PR fixes the error and properly removes all chips.


CodeHeeler (approved):
This fixes the issue, all filters clear chips clear now, thank you so much!

PR #405, opened 2020-12-11
Fix broken tooltips on state sort

johnsonm325:
Steps to reproduce:
1. Create a comparison
2. See that tooltips appear
3. Sort by state and see that some tooltips don't appear

This PR fixes this issue and all tooltips appear on state sort.


petrblaho (approved):
LGTM!

CodeHeeler (approved):
Yes, this fixed the bug and all tooltips remain visible after sorting by state, thank you so much!

PR #406, opened 2020-12-15
Add active global filters

johnsonm325:
This PR adds an alert to the top of the add system and create baseline modals that shows any active global filters. The filter will only show up if you choose a filter. If you only select 'All Workloads' but don't select any tags or SAP IDs, then you won't see the alert.

You should be able to close this alert by clicking the 'X'. If you back out of the modal and go back in, the alert should reappear.

### DIFF
--- a/src/SmartComponents/AddSystemModal/AddSystemModal.js
+++ b/src/SmartComponents/AddSystemModal/AddSystemModal.js
@@ -6,6 +6,7 @@ import { sortable, cellWidth } from '@patternfly/react-table';
 
 import SystemsTable from '../SystemsTable/SystemsTable';
 import BaselinesTable from '../BaselinesTable/BaselinesTable';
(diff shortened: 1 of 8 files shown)


petrblaho (approved):
LGTM!

johnsonm325:
@petrblaho I'll be pinging you again for a re-review of this. We have to add it to the Create Baseline modal when you are creating a baseline from an existing system, as the global filters will apply there as well.

johnsonm325:
@petrblaho pinging for a re-review. Check for the same behavior in the create baseline modal.

Baselines -> click "Create Baseline" button -> select "Copy an existing system" radio button

You should get the same interaction as you do in the add system modal.

petrblaho (approved):
Thank you, @johnsonm325 . It works as described!

PR #411, opened 2021-01-05
Fix formatting of comparison CSV

johnsonm325:


CodeHeeler (approved):
This works, the headers are now taken into account for the values in the columns.  Thanks!

PR #414, opened 2021-01-11
Remove tooltip from comparison export to CSV

johnsonm325:


CodeHeeler (approved):
The tooltip wording has been removed.  Thanks!

PR #415, opened 2021-01-12
add ouia attributes to fact name sorting

opacut:
Adding ouia attributes to the Fact sorting header on Comparison table


johnsonm325 (approved):
Great job @opacut. Looking forward to seeing all of the improvements you bring to testing our app.

PR #416, opened 2021-01-14
Update config and inventory to chrome 2.0 compatible version.

Hyperkid123:
This PR migrates the application build to be compatible with the upcoming version of chrome. This migration is required before we can turn on chrome 2.0 features because of the breaking change in inventory component loading. The application still works with the current version of chrome and until chrome 2.0 is deployed, it still uses the current inventory loading system (which does not work in chrome 2.0).

Changes
- update dependencies required by chrome 2 build config
- update build configurations in preparation for chrome 2.0
- use update chrome 2.0 compatible version of Inventory component

cc @ehelms @johnsonm325 It would be great if you could test out the changes. You know the app much better than I. Also, please check if custom styles are working. Styles are now scoped by drift className to avoid potential global CSS conflicts.

Migration guide: https://github.com/RedHatInsights/insights-chrome/blob/chrome-2.0/docs/migrationGuide.md (does not include inventory)


johnsonm325:
I've found this one bug in the CSS. We've lost the padding to this Form, even though the className remains. I'm not sure why the classNames are there `md-padding-top` and `md-padding-bottom`, but the padding doesn't show.

New:
image
image

Old:
image
image

johnsonm325:
@Hyperkid123 In fact, it looks like a lot of things in that modal are off. Also, some orientation of items in the cells of some of the tables are not where they used to be.

I'm not sure what of this is a problem uncovered by this PR and what is a problem caused by this PR. If you have any insight, please let me know. If you'd like for us to set up a call for "pair programming" that can work as well. I know you're GMT+1, so your 5:00 PM is my 11:00 AM. I can meet between 9:00 AM - 11:00 AM (3 - 5 PM your time) Thursday or Friday.

Hyperkid123:
I know whats the issue. Its because the modal appends to body element but the class that scopes the css is on the react root element. Fortunately we can change the modal portal root. I will adjust it in the morning. If more styling issues surface we can set up a call and deal with them all at once.

Hyperkid123:
@johnsonm325 Turns out that adding `drift` class name to modals is good enough. It's better than changing the default modal root element.

While I was doing it I have noticed that you have quite a lot of custom spacing classes (for example `md-padding-bottom`) that have the exact same value as "official" PF modifiers. I would definitely try to use those since that is the standard. Plus you might remove some classes and you will be covered when UX says "this spacing seems a bit weird to me" :smile: .

Check the docs. I am talking about `pf-u-x-x` set of modifiers. https://www.patternfly.org/v4/utilities/spacing#examples

johnsonm325 (approved):
LGTM! Thanks for the help on this. I appreciate all of your work.

PR #417, opened 2021-01-15
DRFT-78 Add ability to filter on multiple fact names

johnsonm325:
Testing directions are outlined below:

- Test that typing in the fact filter box adds the filter to the filter chips as you are typing.
- Test that pressing enter removes the fact filter from the text box and keeps it in the filter chips.
- Test that adding a filter with the full category name and pressing enter filters that category and all of its sub-facts
- Test that removing a single fact filter chip works.
- Test that removing all the fact filter chips works.
- Test that clearing all filters also removes all of the fact filter chips.
- Test that if you have a fact filter in the text box and you remove that filter chip, it removes it from the text box.
- Test that you can't add 2 of the same fact filters.
- Test that a single fact filter narrows down the facts to only match that filter.
- Test that adding 2 filters gives you results that match both of those filters.
- Removing and adding fact filters should update the facts shown, immediately.


sthirugn:
Can you add `DRFT-78` in the commit message as well?

sthirugn:
@johnsonm325++ thank you for helping me review this PR.

All the scenarios mentioned in the description worked great with the exception of one:
- typing the category name and hitting enter, created a new chip but cleared the result.

sthirugn (changes requested):
All the scenarios mentioned in the description worked great with the exception of one:
* typing the category name and hitting enter, created a new chip but cleared the result.

sthirugn (approved):
/lgtm. It works great now.

PR #423, opened 2021-01-21
DRFT-113 Fix persistent activeFactFilter on new fetch

johnsonm325:
To reproduce:
1. Add a system for comparison
2. Filter by fact.  Works correctly.
3. Add another system.

The facts are no longer filtered, although the filter chip is still there.

This PR fixes the bug where an empty array was being added to the a filter function in the reducer on FETCH_COMPARE_FULFILLED instead of passing the activeFactFilter in the state.


sthirugn:
works awesome.

sthirugn (approved):

PR #424, opened 2021-01-22
DRFT-121 Fix clear all comparisons for activeFactFilters

johnsonm325:
To repro:
Add two systems
Apply filter by name, press enter
Apply another name filter, don't press enter
Clear all comparisons
Add a system

Only one of the name filters remains applied. This PR keeps all filters on clear all comparisons.


CodeHeeler (approved):
Yes, all filters are maintained after clear all comparisons now.  Thanks!

PR #426, opened 2021-01-25
Revert "Update config and inventory to chrome 2.0 compatible version."

johnsonm325:
Reverts RedHatInsights/drift-frontend#416


Hyperkid123:
@johnsonm325 I am looking at the issue. We are dealing with a "common" webpack loader issue when the order CSS modules are different in dev and prod build.

PR #429, opened 2021-01-28
Fix stable build by using new version of config and passing custom deployment

karelhala:
### Fix stable deployments

Since drift is using custom release that releases stable builds from master branch we have to pass in `deployment` config variable to shared config. This has been enabled in most recent version.


karelhala:
ping @johnsonm325

sthirugn (approved):
thank you @karelhala

PR #431, opened 2021-01-29
Check if SAP is selected

karelhala:
### Check if workloads filter has SAP selected

Because workloads has been changed and now it is checkbox we have to check if SAP is really selected and if so show the global filter, otherwise no need to show it at all.


johnsonm325 (approved):
@karelhala thanks for fixing this, and thanks for the tests!

PR #433, opened 2021-02-03
[DRFT-159] Fix workloads: undefined in global filter alert

johnsonm325:
To reproduce:
- Spin up fresh drift page with no pre-selected global filters
- Choose a SAP ID (SID) from the global filters, but do not choose a workload
- Click the `Add to comparison` button to view the modal

You should see `Workloads: undefined`

With this fix, the workloads won't show up if you haven't filtered by a workload

### DIFF
--- a/src/SmartComponents/GlobalFilterAlert/GlobalFilterAlert.js
+++ b/src/SmartComponents/GlobalFilterAlert/GlobalFilterAlert.js
@@ -21,7 +21,10 @@ export class GlobalFilterAlert extends Component {
 
     buildBody = () => {
         const { sidsFilter, tagsFilter, workloadsFilter } = this.props.globalFilterState;
-        let filters = 'Workloads: ' + Object.keys(workloadsFilter)[0] + '. ';
+
+        let filters = Object.keys(workloadsFilter).length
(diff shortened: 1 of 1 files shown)


eherget (approved):
Once I worked through my local dev issues, this works as expected.  Approving.

One note... I got a warning during compile that I didn't see when I ran from master branch.

WARNING in ./src/constants.js 13:22-29
Should not import the named export 'version' (imported as 'version') from default-exporting module (only default export is available soon)
 @ ./src/SmartComponents/modules/actions.js 3:0-44 42:29-32 42:35-39 42:42-45 49:22-25 50:14-18 51:29-33 54:14-17
 @ ./src/App.js 6:0-56 112:15-42 113:15-47 114:15-42
 @ ./src/AppEntry.js 6:0-24 15:38-41
 @ container entry ./RootApp[0]

webpack 5.16.0 compiled with 1 warning in 43138 ms

petrblaho (approved):

johnsonm325:
@eherget Good catch. I found a way to fix this warning. We can `import version from package.json` instead of `import { version } from package.json`. I'm gonna fix that in another PR though.

PR #435, opened 2021-02-04
[DRFT-147] Ouia baselines table

opacut:


src/SmartComponents/BaselinesTable/BaselinesTable.js
```diff
@@ -115,6 +118,7 @@ export class BaselinesTable extends Component {
                     tableId={ tableId }
                     baselineRowData={ baseline }
```
johnsonm325, line 119: This is definitely out of scope with what you are working on, but I like what you did with passing the `baselineName` in instead of the entire row, so can you update BaselinesTable by passing `baselineId={ baseline[0] }` and update BaselineTableKebab by removing references to `baselineRowData` with `baselineId`?

src/SmartComponents/BaselinesTable/BaselineTableKebab/BaselineTableKebab.js
```diff
@@ -38,16 +38,18 @@ export class BaselineTableKebab extends Component {
 
     render() {
         const { isOpen, modalOpened } = this.state;
-        const { baselineRowData, fetchWithParams, tableId } = this.props;
+        const { baselineRowData, fetchWithParams, tableId, baselineName } = this.props;
```
johnsonm325, line 41: `const { baselineId, fetchWithParams, tableId, baselineName } = this.props;`

src/SmartComponents/BaselinesTable/BaselineTableKebab/BaselineTableKebab.js
```diff
@@ -83,7 +89,8 @@ BaselineTableKebab.propTypes = {
     baselineRowData: PropTypes.array,
```
johnsonm325, line 89: `baselineId: PropTypes.string`

johnsonm325 (changes requested):
See my comments about making some additional changes to what's passed from `BaselinesTable` to `BaselineTableKebab`. I called out a couple of specific changes to make to `BaselineTableKebab`, but there are more that you'll need to make to finish it out. Let me know if you need guidance.

opacut:
Added the proposed changes, and rebased branch.
Thank you for the review!

src/SmartComponents/BaselinesTable/BaselinesTable.js
johnsonm325, line 119: You can remove this parameter now that it's not being used by BaselineTableKebab.

johnsonm325 (changes requested):
Small change here.

opacut:
fixed and rebased

johnsonm325 (approved):

PR #436, opened 2021-02-04
[DRFT-165] Fix version import in constants.js

johnsonm325:
We were importing as `{ version }` when the import of the file is an object. We simply had to import `version` from the returned object.

To reproduce:
1. run `npm run start` on master and see the compiler warning.

This fix will cause the compiler to compile successfully every time instead of with warnings.


eherget (approved):
Compile warning is gone now.  LGTM

PR #437, opened 2021-02-05
[DRFT-147] ouia for baseline view

opacut:
- Add fact or category button
- Edit Baseline kebab
- Edit baseline table row, header row
- _category_ attribute for subfacts
- bulk selector - but only the Dropdown element

### DIFF
--- a/src/SmartComponents/BaselinesPage/EditBaseline/EditBaseline.js
+++ b/src/SmartComponents/BaselinesPage/EditBaseline/EditBaseline.js
@@ -175,7 +175,10 @@ export class EditBaseline extends Component {
 
     renderHeaderRow(hasWritePermissions) {
         return (
-            <tr key='edit-baseline-table-header'>
+            <tr
+                key='edit-baseline-table-header'
+                data-ouia-component-type='PF4/TableHeaderRow'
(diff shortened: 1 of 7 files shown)


src/SmartComponents/BaselinesPage/EditBaseline/EditBaseline.js
```diff
@@ -265,6 +276,8 @@ export class EditBaseline extends Component {
 
         return (
             <Checkbox
+                data-ouia-component-type='PF4/Checkbox'
+                data-ouia-component-id={ 'checkbox-' + fact[1] }
```
johnsonm325, line 280: `data-ouia-component-id={ 'checkbox-' + fact[FACT_NAME] }`
`FACT_ID`, `FACT_NAME` and `FACT_VALUE` are constants that are saved in another file. They are 0, 1 and 2, respectively. So instead of using 1 here, you can just use `FACT_NAME`.

johnsonm325 (changes requested):
Minor change here. Otherwise, it looks great

opacut:
fixed and rebased

johnsonm325 (approved):

PR #438, opened 2021-02-05
[DRFT-144] Add filters and sorting to url

johnsonm325:
To test:
- Create a comparison with any number of systems (to test best, add a system, baseline and hsp)
- See the filters and sort in the url
- Modify the url manually to achieve results
- Modify the filters and sorting via normal means (clicking buttons, typing in filters, etc)
- Add and remove systems to ensure that the url properly updates in response to your actions
- Ensure that the page updates in response to your edits to the url

For filtering and sorting, we are using the following rules based on the provided links.
- For filtering rules, see: https://jsonapi.org/recommendations/#filtering
- For sorting rules, see: https://jsonapi.org/format/#fetching-sorting

Here are some rules for how the urls are expected to look:
- Fact name filters are comma separated and follow `filter[name]=`
- State filters are comma separated and follow `filter[state]=`. If they are in the url, they should be selected. If they do not appear in the url, then they should be deselected
- Sorts are comma separated and follow `sort=`. A dash symbol in front of the sort signifies that it is a descending sort. No symbol signifies ascending sort. If fact is not added to the sort, the app will automatically set it to default descending. If state is not included in the sort, it will be set to no sort.

Here is an example of a complete filtering/sorting url:
`?filter[name]=bios,arch&filter[state]=same,incomplete_data&sort=-state,fact`

Here is an example of a full url with system ids and filtering/sorting:
`/insights/drift/?baseline_ids=<baseline_id>&hsp_ids=<hsp_id>&reference_id=<reference_id&system_ids=<system_id>&filter[state]=same,different,incomplete_data&sort=-state,fact`

### DIFF
--- a/src/Utilities/SetHistory.js
+++ b/src/Utilities/SetHistory.js
@@ -1,14 +1,39 @@
 import queryString from 'query-string';
+import { ASC, DESC } from '../constants';
 
-export function setHistory(history, systemIds = [], baselineIds = [], hspIds = [], referenceId) {
+export function setHistory(
+    history, systemIds = [], baselineIds = [], hspIds = [], referenceId, activeFactFilters = [], factFilter, stateFilters, factSort, stateSort
+) {
+    let nameFilters = [ ...activeFactFilters, ...factFilter ? [ factFilter ] : [] ];
+    let filterState = [ ...stateFilters?.filter(({ selected }) => selected)?.map(({ filter }) => filter?.toLowerCase()) || [] ];
+    let sort = [
+        ...[ ASC, DESC ].includes(stateSort) ? [ `${ stateSort === DESC ? '-' : '' }state` ] : [],
+        ...[ ASC, DESC ].includes(factSort) ? [ `${ factSort === DESC ? '-' : '' }fact` ] : []
+    ];
+    let searchPrefix = '?';
+
     /*eslint-disable camelcase*/
     history.push({
-        search: '?' + queryString.stringify({
+        search: searchPrefix + queryString.stringify({
             system_ids: systemIds,
             baseline_ids: baselineIds,
             hsp_ids: hspIds,
             reference_id: referenceId
         })
+    });
+
+    searchPrefix = '&';
+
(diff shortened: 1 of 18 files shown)


katierik:
This looks good to me. -- I'm assuming that if the user enters the URL with a specific filter that filter will always be used. If the user doesn't enter with a specific filter they will get the default.

johnsonm325:
> This looks good to me. -- I'm assuming that if the user enters the URL with a specific filter that filter will always be used. If the user doesn't enter with a specific filter they will get the default.

Yes, if you enters a url with specific filter, will be used until the user clears filters or changes the filters. And, they will get defaults if nothing is entered.

src/Utilities/SetHistory.js
```diff
@@ -1,14 +1,58 @@
 import queryString from 'query-string';
+import { ASC, DESC } from '../constants';
+
+export function setHistory(
+    history, systemIds = [], baselineIds = [], hspIds = [], referenceId, activeFactFilters = [], factFilter, stateFilters, factSort, stateSort
+) {
+    let nameFilters = [ ...activeFactFilters ];
```
karelhala, line 7: This could be simplifiedusing spread operator
```JS
let nameFilters = [ ...activeFactFilters, ...factFilter ? [factFilter] : [] ];
let filterState = [ ...stateFilters?.filter(({ selected }) => selected)?.map(({ filter }) => filter?.toLowerCase()) || [] ];
lest sort = [
  ...[ASC, DESC].includes(stateSort) ? [`${stateSort === DESC ? '-' : ''}state`] : [],
  ...[ASC, DESC].includes(factSort) ? [`${factSort === DESC ? '-' : ''}fact`] : [],
];
```

karelhala (commented):
Just a few code improvements, but I think it looks solid. I didn't checked in overall picture though.

johnsonm325:
@karelhala This is ready for re-review. Thanks for taking a look.

karelhala (approved):
Looking hella good!

eherget (approved):
karelhala says its good!

PR #439, opened 2021-02-08
[DRFT-147] ouia for comparison view

opacut:
- table row, header row
- pagination (top lvl element only)
- category information on subfacts
- expand category button
- drift header cells

### DIFF
--- a/src/SmartComponents/HistoricalProfilesPopover/__tests__/__snapshots__/HistoricalProfilesPopover.tests.js.snap
+++ b/src/SmartComponents/HistoricalProfilesPopover/__tests__/__snapshots__/HistoricalProfilesPopover.tests.js.snap
@@ -532,7 +532,7 @@ exports[`ConnectedHistoricalProfilesPopover should render correctly 1`] = `
           >
             <span
               className="hsp-icon-padding historical-system-profile-popover"
-              data-ouia-component-id="hsp-popover-toggle"
+              data-ouia-component-id="hsp-popover-toggle-undefined"
               data-ouia-component-type="PF4/Button"
             >
               <Popover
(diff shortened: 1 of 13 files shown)


src/SmartComponents/HistoricalProfilesPopover/__tests__/__snapshots__/HistoricalProfilesPopover.tests.js.snap
```diff
@@ -532,7 +532,7 @@ exports[`ConnectedHistoricalProfilesPopover should render correctly 1`] = `
           >
             <span
               className="hsp-icon-padding historical-system-profile-popover"
-              data-ouia-component-id="hsp-popover-toggle"
+              data-ouia-component-id="hsp-popover-toggle-undefined"
```
johnsonm325, line 535: I'm not sure what you want to do here. There are instances where the id is undefined. Do you want to keep `undefined` here? Or do you want to keep it as `hsp-popover-toggle`?

johnsonm325 (changes requested):

src/SmartComponents/HistoricalProfilesPopover/__tests__/__snapshots__/HistoricalProfilesPopover.tests.js.snap
opacut: I didn't know that. What cases can cause an HSP id to be undefined?

opacut: I think we can leave it as undefined for now

johnsonm325 (approved):

PR #442, opened 2021-02-09
ouia for add to comparison modal

opacut:
add to comparison modal
- modal buttons
- tab buttons
- baseline table toolbar
- top-lvl table elements


opacut:
rebased

johnsonm325 (approved):
With updated snapshots and changes to submit and cancel buttons in modal, this is good to go.

PR #444, opened 2021-02-12
Change ouiaId for create baseline modal submit button for Pendo

johnsonm325:
This change will allow us to determine which radio button was selected when a user creates a baseline from the create baseline modal. We will set 3 features in Pendo to track each ouiaId and see how users are creating baselines.

To test, go to the create baselines modal and open the developer tools in your browser. Use the element inspector to select the 'Create Baseline' button (you'll need to type something in the Baseline name input box to get the 'Create baseline' button to be selectable. As you change the selected radio button, you should see the `data-ouia-component-id` update with one of the following strings:

- `copy-baseline-button-fromscratch`
- `copy-baseline-button-copybaseline`
- `copy-baseline-button-copysystem`

### DIFF
--- a/src/SmartComponents/BaselinesPage/CreateBaselineModal/CreateBaselineModal.js
+++ b/src/SmartComponents/BaselinesPage/CreateBaselineModal/CreateBaselineModal.js
@@ -46,6 +46,21 @@ export class CreateBaselineModal extends Component {
         };
     }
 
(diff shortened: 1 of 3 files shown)


CodeHeeler (approved):
Perfect, inspected the element, it behaved exactly as expected.  Thanks!

PR #453, opened 2021-03-05
Add popover with collected selections from add system modal

johnsonm325:
The total number of systems, baselines and hsps that have been selected in the add system modal should be collected at the top right of the add system modal as a link, showing `Selected (2)` or `Selected (0)`.

Upon clicking the link, everything you have selected will show up in a scrollable table with checkboxes. To remove a system, baseline or hsp, simply uncheck the boxes for those you want to remove and upon clicking the 'X' button at the top of the popover, those systems will be deselected in the add system modal. You can select and deselect items in the basket as much as you want but they will not be removed until you close the popover.

To test, try as many ways as possible to get an incorrect number to show up in the link, to get a repeat of a selected item, or to fail to properly add or remove an item from the basket.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [x] File Management
- [ ] Memory Management


CodeHeeler:
Just adding a comment in case anyone else is testing, just fyi that I found a couple things in testing, and will retest once those are addressed, johnsonm325 said likely mid next week.  The popover doesn't disable me mucking around on the original check boxes, and then also navigating pages does not maintain correct Selected count and attempting to then click selected after navigating back made it run away and hide. :)

CodeHeeler:
I broke it :sweat_smile:

but it becomes a question of how terribly impatient a user are you working for maybe?  I'm not sure how likely an average user is to run into it

basically, it works, but there's no way to cancel your selections, like, if you don't click x and just click away from the popover, it goes away but the unclicked item is still removed.  Not sure if that's expected, since yes, the popover closed and then the stuff took effect, but there's no way to go "cancel" without having to recheck anything that's unchecked, not sure if that's something someone would want to do.

but then I took it a step further and really broke it

I had the basket open with five things in it, I unclicked an item, but instead of clicking x or just simply clicking away, I clicked the submit button, all five then showed up in the comparison, including the one I'd unchecked, and now when I go back to add to comparison, the basket is preopened but clicking that same one that should have left won't get rid of it even when then clicking x, and when  I unclick anything else, it now says the number goes down for the basket but when I navigate to those baselines, they're still checked in the regular selection so it's just totally broken.

but again, not sure how likely someone is to do all of that as an average user?

PR #454, opened 2021-03-09
[DRFT-114] Fix state on side nav click

johnsonm325:
Before, when a user clicked 'Comparison' or 'Baseline' in the side nav, we were inadverently clearing the state of redux and if you, for example, created a comparison, then clicked 'Baseline' in the side nav, and then went back to 'Comparison', the comparison you previously had would be cleared out. There was a refresh happening that we didn't realize was happening.

When that refresh was removed, suddenly we had a persistent state. So, in the above example, you now would keep you comparison when you navigated to other parts of the app via the side nav. The problem is, your baseline selections in the add system modal would persist across the 'Comparison' and 'Baseline'. This fix allows your state to remain across side nav clicks, but to have unique system and baseline selections across your side nav clicks.

Primary testing:

- Any baseline, system and hsp selections in the following should remain unique in the following areas:
  - Add system modal (check systems and baselines tabs)
  - Baselines Table
  - Create baseline modal (check create from system and create from baseline radio buttons)

More importantly, the refactors made in this PR can have affected many other parts of the application. I would suggest checking as much as you can. Especially in the Baselines side nav

- Create baseline in every form
- Delete baseline(s)
- Create multiple comparisons and click back and forth in side nav
- Example, create a comparison, then click over to create a baseline from something and then click back and forth
- etc.

We'll mostly rely on QE to really find the bugs. Especially the QA automation runs.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [x] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/AddSystemModal/AddSystemModal.js
+++ b/src/SmartComponents/AddSystemModal/AddSystemModal.js
@@ -10,6 +10,7 @@ import GlobalFilterAlert from '../GlobalFilterAlert/GlobalFilterAlert';
 import { addSystemModalActions } from './redux';
 import { baselinesTableActions } from '../BaselinesTable/redux';
 import { historicProfilesActions } from '../HistoricalProfilesPopover/redux';
+import systemsTableActions from '../SystemsTable/actions';
 
 export class AddSystemModal extends Component {
     constructor(props) {
@@ -28,6 +29,7 @@ export class AddSystemModal extends Component {
 
     async componentDidMount() {
         await window.insights.chrome.auth.getUser();
+        this.props.updateColumns('display_name');
     }
 
     onSelect = (event, isSelected, rowId) => {
@@ -42,7 +44,7 @@ export class AddSystemModal extends Component {
             ids = [ baselineTableData[rowId][0] ];
         }
 
-        selectBaseline(ids, isSelected, 'CHECKBOX');
+        selectBaseline(ids, isSelected, 'COMPARISON');
     }
 
     confirmModal() {
@@ -77,7 +79,7 @@ export class AddSystemModal extends Component {
             ids.push(baseline[0]);
         });
 
-        selectBaseline(ids, isSelected, 'CHECKBOX');
+        selectBaseline(ids, isSelected, 'COMPARISON');
     }
 
     render() {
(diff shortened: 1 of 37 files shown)


CodeHeeler (changes requested):
Ran into some bugs.  While it initially appeared decoupled after starting a comparison and navigating to baselines tab and finding them unset, upon selecting a couple baselines there and navigating back to the comparisons tab, the reference baseline has become unselected, and trying to click anything triggers a brand new add to comparison screen, and when attempting to load the list there, the page crashes. (video shared internally via slack).

CodeHeeler:
The bugs from before are fixed!  I am noticing two things, but am unsure if it's this PR or other instabilities, so here they are in case others can repro, otherwise we'll check again later.
After creating a comparison with some systems, baselines, and hsps, swapping to baselines tab, adding a couple baselines, swapping back, removing a couple things from the comparison, adding to comparison, all looked good. But...
1) If I remove a system from the comparison report, related hsps immediately disappear too, is this expected?
2) If I go to baselines tab, click create baseline, select copy existing system, let the systems load, select one, everything locks up, can't cancel can't click anything, spins and won't show systems any longer, have to kill the browser.

CodeHeeler (changes requested):
Everything appears to work and remain detached now except creating a baseline. I add a name, I select copy an existing system, then whether I select a system or an hsp, the box to create will turn blue instead of grey but clicking it does nothing and I have to cancel to get out of the screen.

Creating a baseline from an existing baseline still works, it's only system or hsp that fails.

CodeHeeler (approved):
OK!  This time I think it works and haven't found any obvious bugs.

The only thing I'm noticing which may be expected behaviour:
- if you create a comparison with two baselines and remove the one that was designated a reference, or if you just uncheck so that there is no reference
- navigate away to the baselines tab
- navigate back and the comparison now designates the first/only baseline as a reference id again automatically

I just can't recall if this is what we want so I'm mentioning it, but everything else is great so I'm approving.

PR #462, opened 2021-04-20
[DRFT-313] Get diff of current and systems to add in notifications

johnsonm325:
open a baseline, system notifications tab
make sure there is at least one system present (system_A)
open Add system
make sure that system_A is visible on the page (with the greyed-out checkmark)
select all via Bulk selector

Clicking Submit does nothing and creates 500 error. This is because bulk select adds all of the systems on the page to the array of systems that are attempted to be added to the baseline notification DB. Because one of those ids is already associated, it fails with a 500. This PR finds the diff of the systems to be added and the existing systems so that a duplicate id isn't requested.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


sthirugn:
I still see 500 when I do the following scenario:
1. I add a new system, submit it.
2. Edit again, without changing anything 

If I do a page refresh between steps 1 and 2, I don't see 500

petrblaho:
Works for me... @sthirugn I am curious what we do differently :-)

johnsonm325:
@sthirugn I was able to repro. Try my new fix.

sthirugn:
I don't see the 500 error any more.

sthirugn (approved):

PR #473, opened 2021-05-11
[DRFT-325] Disable 'Add system' button with no baselinesWrite perms

johnsonm325:
To test:

- Go to line 29 in `src/App.js` and change to:
```
    const handlePermissionsUpdate = (hasCompareRead, hasBaselinesRead/*, hasBaselinesWrite*/, hasInventoryRead) => {
        setPermissions({
            hasCompareReadPermissions: hasCompareRead,
            hasBaselinesReadPermissions: hasBaselinesRead,
            hasBaselinesWritePermissions: /*hasBaselinesWrite*/false,
            hasInventoryReadPermissions: hasInventoryRead,
            arePermissionsLoaded: true
        });
    };
```
- Go to details for a baseline
- Select `Systems` tab

'Add system' button should be disabled. If you revert the above code then you'll see that the button is now enabled.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/AddSystemModal/AddSystemModal.js
+++ b/src/SmartComponents/AddSystemModal/AddSystemModal.js
@@ -173,9 +173,9 @@ export class AddSystemModal extends Component {
 
     render() {
         const { activeTab, addSystemModalOpened, baselineTableData, globalFilterState, handleBaselineSelection, handleHSPSelection,
-            hasBaselinesReadPermissions, hasBaselinesWritePermissions, hasInventoryReadPermissions, hasHSPReadPermissions, historicalProfiles,
-            loading, entities, selectEntity, selectHistoricProfiles, selectedBaselineIds, selectedBaselineContent, selectedHSPContent, selectedHSPIds,
-            selectBaseline, selectedSystemContent, selectedSystemIds, setSelectedSystemIds, totalBaselines } = this.props;
+            historicalProfiles, loading, entities, permissions, selectEntity, selectHistoricProfiles, selectedBaselineIds, selectedBaselineContent,
+            selectedHSPContent, selectedHSPIds, selectBaseline, selectedSystemContent, selectedSystemIds, setSelectedSystemIds,
+            totalBaselines } = this.props;
         const { columns, basketIsVisible } = this.state;
 
         return (
@@ -243,10 +243,10 @@ export class AddSystemModal extends Component {
                         >
                             <SystemsTable
                                 selectedSystemIds={ selectedSystemIds }
-                                hasHistoricalDropdown={ hasHSPReadPermissions }
+                                hasHistoricalDropdown={ permissions.hspRead }
                                 historicalProfiles={ historicalProfiles }
                                 hasMultiSelect={ true }
-                                hasInventoryReadPermissions={ hasInventoryReadPermissions }
+                                permissions={ permissions }
                                 entities={ entities }
                                 selectVariant='checkbox'
                                 onSystemSelect={ setSelectedSystemIds }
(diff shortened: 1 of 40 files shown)


vkrizan (approved):
Code LGTM. Thank you.

I've noticed that the render functions get the each permission as a separate param. They might get all permissions (from `value.permissions`) and then pick from them using a destructor. 

Something like:
```
renderMain({ baselinesWrite, inventoryRead }) {
    ...

# called as renderMain(value.permissions)
```
That way it does not need to provide a new perm each time it needs one. :wink:

johnsonm325:
Ready for another review.

src/SmartComponents/AddSystemModal/AddSystemModal.js
```diff
@@ -243,10 +243,10 @@ export class AddSystemModal extends Component {
                         >
                             <SystemsTable
                                 selectedSystemIds={ selectedSystemIds }
-                                hasHistoricalDropdown={ hasHSPReadPermissions }
+                                hasHistoricalDropdown={ permissions.hspRead }
```
vkrizan, line 246: Observation: `SystemsTable` already gets `permissions`.

src/SmartComponents/BaselinesPage/EditBaselinePage/EditBaseline/EditBaseline.js
```diff
@@ -39,13 +39,13 @@ class EditBaseline extends Component {
         this.fetchBaselineId();
     }
 
-    renderHeaderRow(hasWritePermissions) {
+    renderHeaderRow(baselinesWrite) {
```
vkrizan, line 42: Not sure if this rename was necessary :slightly_smiling_face: .

vkrizan (commented):
Oh, these are lot of changes.

I see that the `permissions` are passed a few levels deep in the function calls. If the last function returns a React component, which I guess most do, then instead of passing the permissions down the line it can grab it from the `PermissionContext` (`<PermissionContext.Consumer>`). I just noticed that the context is/was used in some top level components. It can also be used in the bottom ones (like `AddSystemModal`). That way you can reduce passing the variable/constant.

In functional React components this would be even easier, ... just by calling `const permissions = useContext(PermissionContext)`. But that's another story.

src/SmartComponents/AddSystemModal/AddSystemModal.js
johnsonm325: So, I only use `<PermissionContext.Consumer>` at the top level components because it makes testing a bit easier and my snapshots less verbose.

Regarding `SystemsTable` already having permissions. Before, we passed `hasHistoricalDropdown` always as true from here because when the `SystemsTable` is being rendered from `AddSystemModal` we always show the historical dropdown. With this, I can decide to render the dropdown depending on whether or not the user has permissions to read hsps.

src/SmartComponents/BaselinesPage/EditBaselinePage/EditBaseline/EditBaseline.js
johnsonm325: I renamed all of the permission names for consistency. This was done in multiple places.

src/SmartComponents/AddSystemModal/AddSystemModal.js
vkrizan: Oh, I see the nuance of the testing. With a helper method within tests it could be doable with no so much hassle. Check https://testing-library.com/docs/example-react-context/. If you find this `permissions` passing as a better option, then feel free to keep it as is. 

As for the `hasHistoricalDropdown` it makes sense as it is.

vkrizan:
I've took it for a spin. No issues observed. Feel free to merge.

@johnsonm325 I usually tend to split refactoring and fixing changes into separate commits.

Thanks again.

PR #474, opened 2021-05-13
[DRFT-364] Fixes url clear on side nav Comparison click

johnsonm325:
To repro:

Select 1 or more systems for comparison
Select Comparison from the left panel.  The comparison table is still displayed.  However, the url params are cleared.

This PR correctly sets the url with the params.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/DriftPage/DriftPage.js
+++ b/src/SmartComponents/DriftPage/DriftPage.js
@@ -37,6 +37,12 @@ export class DriftPage extends Component {
         this.props.loadEntities();
     }
 
+    componentDidUpdate(prevProps) {
+        if (prevProps.location.search !== '' && this.props.location.search === '') {
+            this.setHistory();
+        }
+    }
+
     setHistory = () => {
         const { activeFactFilters, baselines, factFilter, factSort, historicalProfiles, history, referenceId, stateFilters,
             stateSort, systems } = this.props;
--- a/src/SmartComponents/DriftPage/__tests__/DriftPage.tests.js
+++ b/src/SmartComponents/DriftPage/__tests__/DriftPage.tests.js
@@ -1,3 +1,4 @@
+/*eslint-disable camelcase*/
 import React from 'react';
 import { shallow, mount } from 'enzyme';
 import { MemoryRouter } from 'react-router-dom';
@@ -31,6 +32,7 @@ describe('DriftPage', () => {
             referenceId: undefined,
             stateFilters: allStatesTrue,
             history: { location: { search: '' }, push: jest.fn() },
+            location: { search: '' },
             clearSelectedBaselines: jest.fn(),
             toggleErrorAlert: jest.fn(),
             clearComparison: jest.fn(),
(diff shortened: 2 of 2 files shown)


src/SmartComponents/DriftPage/DriftPage.js
```diff
@@ -37,6 +37,12 @@ export class DriftPage extends Component {
         this.props.loadEntities();
     }
 
+    componentDidUpdate(prevProps) {
+        if (prevProps.location.search !== '/' && this.props.location.search === '') {
```
vkrizan, line 41: When would the `prevProps.location.search` contain `/`? The query string always start with `?`.

Did you wanted to set the History when there is the transition from some querystring to a cleared one, basically getting the event when the "Comparison" is being clicked from the side nav?

If yes, you can try this condition:
```
prevProps.location.search !== '' && this.props.location.search === ''
```

vkrizan (commented):
Personally I find the nav clearing the search as a feature, rather as a bug. :slightly_smiling_face:  But there might be some use case when this is not desired.

src/SmartComponents/DriftPage/DriftPage.js
johnsonm325: > Personally I find the nav clearing the search as a feature, rather as a bug. slightly_smiling_face But there might be some use case when this is not desired.

This is something that we want. We want to keep the comparison if you navigate away to Baselines to do some stuff and then come back.

> When would the `prevProps.location.search` would be `/`? The query string always start with `?`.

If you are on the Comparison nav and then click the Comparison nav again, `props.location.search` is `/`. This function works because I can catch this very specific scenario and revert.

vkrizan: > This is something that we want. We want to keep the comparison if you navigate away to Baselines to do some stuff and then come back.

I see, make sense.

> If you are on the Comparison nav and then click the Comparison nav again, props.location.search is /. This function works because I can catch this very specific scenario and revert.

The first condition there is "not equal" `prevProps.location.search !== '/'` which would be always true, as `search` (query string) is either empty or starts with `?`. The condition is then redundant.

When clicked on the nav:
```
# prevProps.location
Object { 
  search: "?system_ids=c7ec8791-8534-41ce-a905-f7ed79c0f5db&filter[state]=different,incomplete_data&sort=-state,fact", 
  pathname: "/", 
  hash: "", 
  key: "i5ov76"
}

# this.props.location
Object { pathname: "/", search: "", hash: "", state: undefined, key: "z8ewez" }
```

johnsonm325: Good catch! All fixed now.

vkrizan: Is it possible to keep it within the limits of the transition (from some query string to none, i.e. clicking on the nav bar) with the condition noted above https://github.com/RedHatInsights/drift-frontend/pull/474#discussion_r636819389? The `componentDidUpdate` is called on any param update. If there would be no query string and an other param is changed, then the `setHistory` will be called each time. I'm not sure if calling `setHistory` several times with the same attributes would cause any problems.

johnsonm325:
okay, fixed. I accidentally added your recommendations from the other PR to this one. So once you've approved this one, I'll handle any rebasing and merge that one. Thanks for the insight.

vkrizan:
@johnsonm325 The fix for the URL looks good. Would you mind taking out the accidental changes? It would be easier to review them separately. :slightly_smiling_face: 

```
git reset HEAD^
git add -p
# select only the URL fix
git commit
```

Thank you.

vkrizan (approved):
Awesome!

PR #497, opened 2021-08-03
[DRFT-608] Make filters case-insensitive

johnsonm325:
To repro:

- Create a comparison
- Use the fact name filter to filter on a fact name, but use uppercase letters instead of lowercase letters.

This should cause the fact you were trying to filter to not show up when you think it should. Now, filters are case-insensitive, so no matter if you use upper or lowercase in the filter or in the fact name, the fact name will show up.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/modules/__tests__/helpers-tests.js
+++ b/src/SmartComponents/modules/__tests__/helpers-tests.js
@@ -1,8 +1,8 @@
 import helpers from '../helpers';
 import stateFiltersFixtures from './state-filter.fixtures';
-import { compareReducerPayloadWithCategory, compareReducerPayloadWithMultiFact,
-    sortedPayloadWithMultiFactAscDesc, sortedPayloadWithMultiFactAscAsc } from './reducer.fixtures';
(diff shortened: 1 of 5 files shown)


jpramos123 (approved):
Worked as expected! Thanks for this @johnsonm325

src/SmartComponents/modules/actions.js
```diff
@@ -70,7 +70,7 @@ function addStateFilter(filterData) {
 function filterByFact(filter) {
     return {
         type: types.FILTER_BY_FACT,
-        payload: filter
+        payload: filter.toLowerCase()
```
vkrizan, line 73: I'd recommend to move the lowercasing to the reducer, if that's possible. The payload would stay original and the logic would be closer.

johnsonm325:
@vkrizan New changes should work. Thanks!

vkrizan (approved):

PR #501, opened 2021-08-18
[DRFT-539] Add scrollbar to top of comparison table

johnsonm325:
The only horizontal scrollbar on the comparison table is located at the bottom of the table, therefore scrolling is not optimal when a user has to scroll all the way to the bottom to scroll horizontally. This PR adds a scrollbar to the top so a user can scroll horizontally from the top or the bottom of the page.

This PR does not fully fix the issue as the end goal is to make the top scrollbar sticky at the top of the page so you can scroll horizontally even if you are in the middle of the page where neither scrollbar is visible. Because we are also attempting to make the header row of the table sticky, we are going to implement the stickiness with that PR.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/DriftPage/DriftTable/DriftTable.js
+++ b/src/SmartComponents/DriftPage/DriftTable/DriftTable.js
@@ -34,8 +34,24 @@ export class DriftTable extends Component {
         this.setReferenceId();
         this.setFilters();
         this.setSort();
+        this.topScroller = React.createRef();
+        this.bottomScroller = React.createRef();
+        this.doubleScroll = this.doubleScroll.bind(this);
         this.fetchCompare = this.fetchCompare.bind(this);
         this.removeSystem = this.removeSystem.bind(this);
+    }
+
+    doubleScroll() {
+        let wrapper1 = this.topScroller.current;
+        let wrapper2 = this.bottomScroller.current;
+
+        wrapper1.onscroll = function() {
+            wrapper2.scrollLeft = wrapper1.scrollLeft;
+        };
+
+        wrapper2.onscroll = function() {
+            wrapper1.scrollLeft = wrapper2.scrollLeft;
+        };
     }
 
     async componentDidMount() {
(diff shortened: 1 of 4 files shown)


jpramos123 (approved):
LGTM!

Scroll was added at the top of the table and I was able to scroll horizontally!

I agree that the scroll should not disappear when the user is at the middle of the table!! 

Thanks

src/SmartComponents/DriftPage/DriftTable/DriftTable.js
```diff
@@ -38,6 +38,19 @@ export class DriftTable extends Component {
         this.removeSystem = this.removeSystem.bind(this);
     }
 
+    doubleScroll() {
+        let wrapper1 = document.getElementById('top-scroll');
+        let wrapper2 = document.getElementById('table-wrapper');
```
vkrizan, line 43: React has a ref support on components that allow to access variable-assigned components. Read more here https://reactjs.org/docs/refs-and-the-dom.html

This approach does not allow to have more instances. It expects unique ids. If a raw JS approach is needed then a relative element find would be more proper. Why not leverage what the React framework offers :slightly_smiling_face:.

vkrizan (commented):
Check the comment for a React-y way of doing this...

... I guess there might be some other lib that supports syncing the scroll.

johnsonm325:
@vkrizan check out the changes and let me know what you think.

vkrizan:
@johnsonm325 The `#comparison-table` can also use the ref to get the width. Otherwise looks good. Thanks!

Generally, I'd avoid using `id`s unless they are globally unique on a rendered page.

PR #547, opened 2021-11-24
[DRFT-912] Fix obfuscated cell shading

johnsonm325:
To repro:
Create a comparison with at least 1 system with obfuscated values and a ensure that the fact that includes obfuscation results in a state of 'DIFFERENT' for the row. The easiest way to do this is create an empty baseline and compare the baseline to a system with obfuscation and a system with a value that isn't obfuscated.
Set the baseline as reference (should happen out of the box)

The obfuscated value (check for fqdn or ipv) should be completely gray with a lock icon, but the value set as 'DIFFERENT' will have a red border on the left and the background of the cell is gray. With this PR, the background of the 'DIFFERENT' cell will be red.

There is some discussion about whether or not we would highlight these values as red because ip addresses and fqdn should not be the same as others, and therefore should not be flagged with 'DIFFERENT'. In this PR, we are not testing the correctness of this, but simply that IF a fact returns the state of 'DIFFERENT' when compared with a system with obfuscation, that we properly shade it red.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/DriftPage/DriftTable/DriftTable.js
+++ b/src/SmartComponents/DriftPage/DriftTable/DriftTable.js
@@ -463,7 +463,7 @@ export class DriftTable extends Component {
                         data-ouia-component-type='PF4/TableRow'
                         data-ouia-component-id={ 'comparison-table-row-' + comparison.name }
                         category={ fact.name }
-                        className={ comparison.state === 'DIFFERENT' ? 'unexpected-row' : '' }>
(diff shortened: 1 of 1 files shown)


jpramos123 (approved):
LGTM

PR #560, opened 2022-01-31
DRFT-984: Add baseline fact filter to dropdown

johnsonm325:
This PR adds a mini-feature that allows you to filter the comparison table by fact type (ie. all facts or baseline facts only). This means that when at least one baseline is present in your comparison, you can filter facts so that only facts that are present in your baselines actually show up in the table results. If a fact is only present in your systems or HSPs, those facts will not show up.

This filter should work in conjuction with existing filters. Also, by selecting "All facts" you will automatically deselect "Baseline facts only", and vice versa. To test:

Go to drift app
Click "Add to comparison" button (ensure you have at least one system or baseline)
Select at least one system and one baseline to create a comparison and click "Submit"

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/App.scss
```diff
@@ -465,3 +465,9 @@
     text-align: left;
   }
 }
+
+.comparison-filter-input-dropdown-width {
```
mkholjuraev, line 469: I am not sure if we really need a static width?

src/Utilities/SetHistory.js
```diff
@@ -2,10 +2,12 @@ import queryString from 'query-string';
 import { ASC, DESC } from '../constants';
 
 export function setHistory(
-    history, systemIds = [], baselineIds = [], hspIds = [], referenceId, activeFactFilters = [], factFilter, stateFilters, factSort, stateSort
+    history, systemIds = [], baselineIds = [], hspIds = [], referenceId, activeFactFilters = [], factFilter, factTypeFilters, stateFilters, factSort,
+    stateSort
 ) {
     let nameFilters = [ ...activeFactFilters, ...factFilter && !activeFactFilters.includes(factFilter) ? [ factFilter ] : [] ];
     let filterState = [ ...stateFilters?.filter(({ selected }) => selected)?.map(({ filter }) => filter?.toLowerCase()) || [] ];
+    let filterFactType = [ ...factTypeFilters?.filter(({ selected }) => selected)?.map(({ filter }) => filter?.toLowerCase()) || [] ];
```
mkholjuraev, line 10: I believe the same result can be achieved without array destruction. I have checked as following:

let factTypeFilters = [{ id: 1, selected:true, filter: 'a'}, {id:2, selected:false,  filter: 'b'}];
const result = factTypeFilters?.filter(({ selected }) => selected)?.map(({ filter }) => filter?.toLowerCase()) || [];
console.log(result);

src/SmartComponents/DriftPage/FilterDropDown/FilterDropDown.js
```diff
@@ -17,77 +16,57 @@ class FilterDropDown extends Component {
...
         return (
             <React.Fragment>
                 <Dropdown
-                    ouiaId='state-filter-dropdown'
+                    ouiaId={ `${ (type.split(' ').join('-')).toLowerCase() }-filter-dropdown` }
                     toggle={ <DropdownToggle
                         onToggle={ this.onToggle }
-                        ouiaId='state-filter-dropdown-toggle' >
-                                    View: { selectedViews }
+                        ouiaId={ `${ (type.split(' ').join('-')).toLowerCase() }-filter-dropdown-toggle` } >
```
mkholjuraev, line 65: I have some code repetition on line 62 and 65. It can be unified with a single const string. But this is really minor you can  ignore this if you do not think important

mkholjuraev (approved):
I have left a few suggetions. Looks good to me. I have run the PR locally, it works as described in the description.

I think one more reviewer would be better as it is a huge PR. Honestly, it was a bit hard to review long code parts.

src/App.scss
johnsonm325: I guess it's a matter of uniformity. If it don't set it, then the filter box size jumps around and I don't like the way it looks. Not sure what you think about this.

MichalSajdik (approved):
LGTM

PR #564, opened 2022-02-12
[DRFT-302] Make comparison header row sticky on scroll

johnsonm325:
This PR makes the comparison table header and top scroll bar sticky to the top of the page.

To test, add various numbers of systems/baselines (specifically systems with facts so you can expand the facts and test the scroll).
The header of the table and top scrollbar should stick to the top of the page, and when you scroll right, the table contents should hide under the left two columns like they normally do.
Testing adding lots and a system or two to make sure that sizing remains in check. I had to split the header out of the table body. So I essentially had to create two tables and then use CSS to ensure that cell sizes update the same.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/DriftPage/DriftTable/ComparisonHeader/ComparisonHeader.js
+++ b/src/SmartComponents/DriftPage/DriftTable/ComparisonHeader/ComparisonHeader.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { Tooltip } from '@patternfly/react-core';
 import { ClockIcon, TimesIcon, ExclamationTriangleIcon, ServerIcon, BlueprintIcon } from '@patternfly/react-icons';
 import { LongArrowAltUpIcon, LongArrowAltDownIcon, ArrowsAltVIcon } from '@patternfly/react-icons';
+import { Skeleton, SkeletonSize } from '@redhat-cloud-services/frontend-components';
 import moment from 'moment';
 
 import { ASC, DESC } from '../../../../constants';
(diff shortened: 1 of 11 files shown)


src/SmartComponents/DriftPage/DriftTable/ComparisonHeader/ComparisonHeader.js
```diff
@@ -47,6 +48,13 @@ class ComparisonHeader extends Component {
         setHistory();
     }
 
+    renderLoadingSystems() {
+        let row = [];
```
mkholjuraev, line 52: I would suggest just simple one liner to achieve the same purpose:

`return [<td><Skeleton size={ SkeletonSize.md } /></td>]`

mkholjuraev (changes requested):
The code looks good to me. The sticky header is working, but I have observed a side effect in the Global header. This needs to be addressed I believe.


image

johnsonm325:
> The code looks good to me. The sticky header is working, but I have observed a side effect in the Global header. This needs to be addressed I believe.
> 
> image

This seems to be a new design. If you go to console.stage.redhat.com, and go to any apps, you'll see they removed the padding. Looks like they switched out a class that did that.

johnsonm325:
@mkholjuraev I've got to fix an issue with the top scroll bar. I may talk to you about this tomorrow, but it appears one way in Firefox and another way in Chrome. I want to find the best fix, but not sure how to do it yet.

mkholjuraev:
@johnsonm325 yes, I agree that the issue is on Platform side. You do not have to fix this, I believe.

johnsonm325:
@mkholjuraev Could you do another review. The only changes I've made is the one-liner you suggested and I added some CSS magic to the top scrollbar. Test the scrollbar in Chrome and Firefox to make sure it works well.

mkholjuraev (approved):
LGTM! I run the PR locally. It is working as expected

PR #578, opened 2022-03-31
[DRFT-236] Sort on baseline facts

MichalSajdik:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


johnsonm325 (changes requested):
I finally got some feedback on this.

1. Regarding the icons next to `Fact` and `Value`, when one column sort is "active", turn off the icon (set to `ArrowsAltVIcon`) for the other column. In the case where we sort based on `Value`, all of the "categories" will show up in the same area. For this, we will implement a secondary sort on the `Fact` column that matches the sorting of `Value` at that time, but the arrow will still remain `ArrowsAltVIcon`.
2. Regarding how we sort numbers, just leave the default string sort as is.

Let me know if you need clarification and we can jump on a call.

MichalSajdik:
Sorting should work based on new specification

johnsonm325 (changes requested):
Everything looks good. The only thing that needs to change is that the Fact sorting should toggle ASC <-> DESC. Only the Value column sort should toggle ASC -> DESC -> OFF.

The only time the Fact column sort should be turned off is if the Value column sort is turned on. Hopefully that makes sense. Let me know if you need more clarification.

MichalSajdik:
I think I understood it, yea it was not clear from the beginning. Hopefully it is alright now

johnsonm325 (changes requested):
This works like a charm. There is one series of changes I would request you make, and I'm sorry I didn't mention this earlier, but for sorting the Fact column, you are using 'factSort', but I think we need to change `fact` to `name`. If you do a search of `factSort` in the app, you'll notice we use `factSort`for sorting the comparison table.

I want to avoid any confusion and potential for redux calls to trip over each other.

johnsonm325 (approved):
This looks great. Thanks for taking care of this and dealing with all of my reviews. :)

PR #579, opened 2022-03-31
[DRFT-233] Update bulk selectors

johnsonm325:
Bulk selectors have been updated in 2 places for Drift:

- Baseline main page - Missing "Select page (#)" and counts for select none. We have removed 'Select all' from baseline list bulk select because there are no use cases that would utilize a full select of the baselines list and adding the ability to select all would cause some logic that would be a bit too complex for what we hope to achieve.
- Baseline detail page (If you click on a specific baseline to see its facts) - Missing counts for select none & select all

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/BaselinesTable/BaselinesTable.js
+++ b/src/SmartComponents/BaselinesTable/BaselinesTable.js
@@ -27,15 +27,22 @@ export class BaselinesTable extends Component {
         super(props);
 
         this.state = {
-            sortBy: {
-                index: 1,
-                direction: 'asc'
+            params: {
+                sortBy: {
+                    index: 1,
+                    direction: 'asc'
+                },
+                search: undefined,
+                orderBy: 'display_name',
+                orderHow: 'ASC',
+                page: 1,
+                perPage: 20
             },
-            search: undefined,
-            orderBy: 'display_name',
-            orderHow: 'ASC',
-            page: 1,
-            perPage: 20
+            bulkSelectType: '',
+            errorMessage: [ 'The list of baselines cannot be displayed at this time. Please retry and if',
+                'the problem persists contact your system administrator.',
+                ''
+            ]
         };
     }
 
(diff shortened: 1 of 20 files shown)


src/SmartComponents/BaselinesTable/BaselinesTable.js
```diff
@@ -27,15 +27,22 @@ export class BaselinesTable extends Component {
...
+                page: 1,
+                perPage: 20
             },
-            search: undefined,
-            orderBy: 'display_name',
-            orderHow: 'ASC',
-            page: 1,
-            perPage: 20
+            bulkSelectType: '',
+            errorMessage: [ 'The list of baselines cannot be displayed at this time. Please retry and if',
```
mkholjuraev, line 42: Do we need to keep error message always in the table component? Maybe it is better to keep error flag/API response status in the params, message in some constants/messages file. Then map the error with corresponding message when an error happened.

mkholjuraev (changes requested):
It seems there is an issue with Baseline page bulk select. When I select a page and proceed to next page, hit page selection again. Then come back to the first page and hit **Select none(0)**, I can see that there are still some systems selected left. I was expecting to unseleact all the systems, regardless of the page. 

https://user-images.githubusercontent.com/59481011/163354686-7116448b-1b2b-48e3-a5bb-7e656b57350c.mp4

src/SmartComponents/BaselinesTable/BaselinesTable.js
johnsonm325: Do you mean to simply save the error message as a constant in another constants file and then import the error message when needed?

I'm fine to do that, but only this component uses the error message, which is why I left it here.

I don't care either way though.

johnsonm325:
@mkholjuraev I fixed the issue you recorded with select none.

Also, there is a new interaction with the bulk select based on feedback from UX. Clicking the bulk select checkbox on a given page should

- select the entire page if nothing on that page is selected.
- deselect all baselines if ANY baseline is selected on that page.

src/SmartComponents/BaselinesTable/BaselinesTable.js
mkholjuraev: In that case, it makes sense

mkholjuraev (approved):
LGTM thanks! Bulk select is working as expected on all pages.

PR #584, opened 2022-07-18
[DRFT-233] Fix bugs with bulk select

johnsonm325:
Systems table bulk select now has "Deselect page (<number>)" to deselect systems that are selected on that page. That same deselect has been added to the notifications systems table on the baseline detail page.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/SystemsTable/systemsHooks.js
+++ b/src/SmartComponents/SystemsTable/systemsHooks.js
@@ -1,0 +1,107 @@
+import React, { useState } from 'react';
+import helpers from '../helpers';
+import api from '../../api';
+import { Spinner } from '@patternfly/react-core';
+
+const useSystemsBulkSelect = (
+    id = '',
+    entities = {},
+    selectEntities,
+    bulkSelectDisabled,
+    isAddSystemNotifications
+) => {
+    console.log(id, 'id');
+    const [ isLoading, setIsLoading ] = useState(false);
+
+    const setTitle = () => {
+        let title;
+        title = <React.Fragment>
+            <Spinner size="sm" />
+        </React.Fragment>;
+
+        return { children: [ title ]};
+    };
+
+    const findChecked = () => {
+        if (entities && entities.selectedSystemIds) {
+            return helpers.findCheckedValue(entities?.total, entities?.selectedSystemIds.length);
+        } else {
+            return null;
+        }
+    };
+
+    const checked = findChecked();
+
+    const onBulkSelect = async (event) => {
+        let toSelect = [];
+        switch (event) {
+            case 'none': {
+                toSelect = { id: 0, selected: false, bulk: true };
+                break;
+            }
+
+            case 'page': {
+                toSelect = { id: 0, selected: true };
+                break;
+            }
+
+            case 'all': {
+                if (isAddSystemNotifications) {
+                    setIsLoading(true);
+                    let systemIds = await api.systemFetch(entities?.total);
+                    setIsLoading(false);
+                    toSelect = { id: 0, selected: true, bulk: true, systemIds };
+                } else {
+                    toSelect = { id: 0, selected: true, bulk: true };
+                }
+
+                break;
+            }
+        }
+
+        console.log(entities, 'entities');
+        //console.log(toSelect, 'toSelect');
+        //console.log(selectEntities, 'selectEntities');
+
+        selectEntities(toSelect);
+    };
+
+    const getItems = () => {
+        return [{
+            title: `Select none (0)`,
+            onClick: () => {
+                onBulkSelect('none');
+            }
+        }, {
+            title: `Select page (${ entities?.count || 0 })`,
+            onClick: () => {
+                onBulkSelect('page');
+            }
+        }, {
+            title: `Select all (${ entities?.total })`,
+            onClick: () => {
+                onBulkSelect('all');
+            }
+        }];
+    };
+
+    const bulkCheckboxClick = () => {
+        if (entities?.selectedSystemIds?.length) {
+            onBulkSelect('none');
+        } else {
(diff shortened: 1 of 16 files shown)


adonispuente (changes requested):
Im not sure if Drift ever got a SPUR audit, but I dont think a Deselect Button is supposed to be implemented like this in the tables. I have to find the table audit that Stefan created, but I for bulk selectors it was [Select None, Select Page, Select All]. The only 'deselect' was manual 1 by 1, pressing the actual square to deselect all, or changing 'select all' to 'deselect all' once all items were selected.

mkholjuraev:
> Im not sure if Drift ever got a SPUR audit, but I dont think a Deselect Button is supposed to be implemented like this in the tables. I have to find the table audit that Stefan created, but I for bulk selectors it was [Select None, Select Page, Select All]. The only 'deselect' was manual 1 by 1, pressing the actual square to deselect all, or changing 'select all' to 'deselect all' once all items were selected.

agree. I also think that we should not have a 'Deselect page' in bulk select options for consistency.

johnsonm325:
good call. not sure why that was implemented on my end. I can make the change easy enough.

johnsonm325:
@adonispuente @mkholjuraev want to take another look? It's been overhauled quite a bit.

mkholjuraev:
@johnsonm325  Taking a look.

src/SmartComponents/SystemsTable/systemsHooks.js
```diff
@@ -0,0 +1,104 @@
...
+
+    const setTitle = () => {
+        let title;
+        if (isLoading) {
+            title = <React.Fragment>
+                <Spinner size="sm" />
+            </React.Fragment>;
+        }
+
+        return { children: [ title ]};
```
mkholjuraev, line 23: I believe this function returns `{ children: [undefined] }` when isLoading is false. Do you think that this will not lead to problems in the future?

src/api.js
```diff
@@ -138,6 +143,22 @@ function deleteSystemNotifications(baselineId, systems) {
...
 }
 
+async function systemFetch(count) {
+    let page = 1;
+    let totalSystems = [];
+    let systems;
+    let systemIds;
+    while (Math.ceil(count / 100) >= page) {
+        let path = `/hosts?per_page=100&page=${page}&order_by=updated&order_how=DESC&staleness=fresh&staleness=stale`;
+        systems = await getSystems(path);
```
mkholjuraev, line 153: This does work. But I would suggest initializing API request calls in a loop and then providing those function call references into Promis.all. This will improve the time to fetch those systems as the requests are triggered almost at the same time instead of waiting for each request.

src/SmartComponents/SystemsTable/systemsHooks.js
johnsonm325: I changed it round a bit to only call this function when `isLoading === true`.

src/api.js
johnsonm325: I've applied this. thanks for the info.

johnsonm325:
@mkholjuraev @adonispuente either of you can review this again.

src/SmartComponents/helpers.js
```diff
@@ -113,10 +113,21 @@ function downloadHelper(baselineData) {
     downloadFile(file, filename, baselineData.type);
 }
 
+function findSelectedCountOnPage(rows) {
+    let selected;
```
mkholjuraev, line 117: Minor: 
```suggestion
    const selected = rows ? rows.filter(row => row.selected) ? [];
```

src/store/reducers.js
```diff
@@ -126,7 +126,12 @@ function selectedReducer(
                         let ids = state.rows.map(function (item) {
                             return item.id;
                         });
-                        selectedSystemIds = [ ...new Set(selectedSystemIds.concat(ids)) ];
+                        if (action.payload.systemIds) {
```
mkholjuraev, line 129: How about this?
```suggestion
                        selectedSystemIds =  [ ...new Set(selectedSystemIds.concat(action.payload.systemIds || ids)
```

mkholjuraev (changes requested):
The bulk select options look consistent with other apps. But when I run the PR and open comparison modal, I get the bulk select dropdown in the systems table disabled even if there are rows in the table.

johnsonm325:
> The bulk select options look consistent with other apps. But when I run the PR and open comparison modal, I get the bulk select dropdown in the systems table disabled even if there are rows in the table.

@mkholjuraev This is the expectation from UX. You can't run a comparison with more than 50 systems, so we don't allow it.

src/store/reducers.js
```diff
@@ -126,7 +126,14 @@ function selectedReducer(
                         let ids = state.rows.map(function (item) {
                             return item.id;
                         });
-                        selectedSystemIds = [ ...new Set(selectedSystemIds.concat(ids)) ];
+
+                        selectedSystemIds = [ ...new Set(selectedSystemIds.concat(action.payload.systemIds || ids)) ];
+                        /*if (action.payload.systemIds) {
```
mkholjuraev, line 131: It seems this is a leftover.

mkholjuraev (commented):
There are a few more things and then we are good to go:
1. When I try to add associated systems to a baseline and open the modal, select some systems, and click on the 'Submit' button, I see that selected system ids are not passed into the API query.
2.  When a table is empty, it is good to disable the bulk select dropdown.

PR #588, opened 2022-07-20
Fix cell width on comparison table

johnsonm325:
Before, if you only compared a couple of systems, the header cell width and the body cell widths did not match. This PR fixes that. You should be able to resize the window and they stay the same width.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


gkarat (commented):
(summing up what we had in the slack discussion)

Are there any ways how could we achieve the same result? Because while running it locally, I noticed that the table had been re-rendered with each `resize` event, which is very ineffective (try to run react profiler with the patch and compare to the current implementation at stage/production) 🤔.

I think it is still doable to do with CSS, but we have to make a trade-off: define the fixed width for each column. However, I don't see any possible visual/usability problems with that: after all, the columns will be equal in width, and the slider will be there to help users scroll to the left/right.

Perhaps, this could be one of the solutions in that case (class names could be wrong, but you get the point):
```
.drift-header.right-border.system-header.sticky-header, .comparison-cell {
  min-width: 350px;
  max-width: 350px;
}
```

src/SmartComponents/DriftPage/DriftTable/ComparisonHeader/ComparisonHeader.js
```diff
@@ -14,6 +14,22 @@ import ReferenceSelector from '../ReferenceSelector/ReferenceSelector';
...
+
+    setColumnWidth = () => {
+        if (this.columnWidth) {
+            this.props.setColumnHeaderWidth(this.columnWidth.current.offsetWidth);
+            this.setState({ refState: this.columnWidth });
+        }
+    };
+
+    componentDidMount() {
+        window.addEventListener('resize', this.setColumnWidth);
```
gkarat, line 32:
```suggestion
        window.addEventListener('resize', debounce(this.setColumnWidth, 500));
```

src/SmartComponents/DriftPage/DriftTable/ComparisonHeader/ComparisonHeader.js
```diff
@@ -239,7 +261,9 @@ ComparisonHeader.propTypes = {
     setHistory: PropTypes.func,
     selectedHSPIds: PropTypes.array,
     selectHistoricProfiles: PropTypes.func,
-    selectedBaselineIds: PropTypes.array
+    selectedBaselineIds: PropTypes.array,
+    columnWidth: PropTypes.number,
+    setColumnHeaderWidth: PropTypes.func
```
gkarat, line 266: Since you are calling the function at the line 27, the function is supposed to be required by the component? If so, could you please add `.isRequired` to the prop type?

gkarat (approved):
Just a minor comment, otherwise the code looks good! I wasn't able to run locally because API requests to api/drift/v1/comparison_report failed with 400. It's been writing to me " Error parsing headers: 'limit request headers fields size'." Can we do something with it now?

PR #593, opened 2022-09-30
fix(HistoricalProfilesPopover): DRFT-836 - Disable compare button when selection is empty

tompsota:
Compare button is now disabled when HSP selection is empty

resolves https://issues.redhat.com/browse/DRFT-836

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/SmartComponents/DriftPage/DriftTable/__tests__/__snapshots__/DriftTable.tests.js.snap
```diff
@@ -11385,15 +11385,7 @@ exports[`ConnectedDriftTable should render systems, baselines and historicalProf
...
-                                        footerContent={
-                                          <Button
-                                            onClick={[Function]}
-                                            ouiaId="hsp-popover-compare"
-                                            variant="primary"
-                                          >
-                                            Compare
-                                          </Button>
-                                        }
+                                        footerContent={null}
```
bastilian, line 11388: There should probably also be a test to verify that the button does show when there is `historicalData` in the state.

tompsota:
@johnsonm325 Okay, agreed.
I accidentally removed my changes when I was modifying the commits, it should be fixed now.

Although in jira it is mentioned that we should disable the button, I think that it looks better when the button is removed completely, what do you think about it?

johnsonm325:
@tompsota I think if a popover has an action we have to show the button disabled so the user knows that there is something to do there when historical profiles do show up.

In other words, I think it's something we can't avoid.

PR #607, opened 2022-11-22
fix(redux)fixing store with hooks

Fewwy:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/AppEntry.js
+++ b/src/AppEntry.js
@@ -1,17 +1,32 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import { BrowserRouter as Router } from 'react-router-dom';
 import { Provider } from 'react-redux';
 import { init } from './store';
 import App from './App';
+import logger from 'redux-logger';
 
 import getBaseName from './Utilities/getBaseName';
 
-const Drift = () => (
-    <Provider store={ init().getStore() }>
-        <Router basename={ getBaseName(window.location.pathname) }>
-            <App/>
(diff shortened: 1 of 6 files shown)


src/AppEntry.js
```diff
@@ -1,17 +1,32 @@
...
-            <App/>
-        </Router>
-    </Provider>
-);
+const Drift = () => {
+    const [ registry, setRegistry ] = useState();
+    const store = registry?.getStore();
+
+    useEffect(() => {
+        setRegistry(IS_DEV ? init(logger) : init());
```
bastilian, line 15: the IS_DEV is irrelevant here. The inventory has this as a webpack build variable and will most likely not work here.

src/AppEntry.js
```diff
@@ -1,17 +1,32 @@
...
+        return () => {
+            setRegistry(undefined);
+        };
+    }, []);
+
+    return (
+
```
bastilian, line 22: You might also need a context and context provider for the registry, in case any of the child components do use it, for example the Inventory might call `getRegistry.register` in it's onload, which would then receive the wrong registry/store. 

The Inventory has an example of how to build and use the context/provider

PR #611, opened 2022-11-30
fix(Comparison page): Fix the header layout

gkarat:
Fixes the header layout for the Comparison page (/insights/draft). Before it took extra space around "Comparison" which looked differently from other pages.

изображение


bastilian (approved):
Thank you @gkarat!

PR #615, opened 2022-12-05
[RHIF-104] Enable app-wide Hotjar text suppression

leSamo:
Related to [RHIF-104](https://issues.redhat.com/browse/RHIF-104).

We would like to implement Hotjar app-wide text suppression for all of our apps, but first we would like to test it on one of the apps and see if it creates any problems. Hotjar is a tool used track user activity on page for debugging and analytics purposes. This suppression is used to hide sensitive information on the page, like system names, etc. To suppress an element according to [Hotjar documentation](https://help.hotjar.com/hc/en-us/articles/115012439167#specific-element-suppression) `data-hj-suppress` attribute should be added to an element making it and all of its children to use text suppression (text will be replaced by asterisks). `data-hj-suppress` is added to the root element so that it's enabled for the whole app.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Authentication and Password Management
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Error Handling and Logging
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/src/SmartComponents/BaselinesPage/BaselinesPage.js
+++ b/src/SmartComponents/BaselinesPage/BaselinesPage.js
@@ -126,7 +126,7 @@ export class BaselinesPage extends Component {
                                     <PageHeaderTitle title='Baselines'/>
                                 </PageHeader>
 
-                                <Main store={ registryContextValue?.registry?.getStore() }>
+                                <Main store={ registryContextValue?.registry?.getStore() } data-hj-suppress>
                                     { value.permissions.baselinesRead === false
                                         ? <EmptyStateDisplay
                                             icon={ LockIcon }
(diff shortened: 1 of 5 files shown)


gkarat (changes requested):
When running locally, I couldn't find any of the Drift components that would have this attribute (just watched HTML through the Inspector window). I would expect the parent container should have this value.

Also please could you write a shallow test that would check for the presence of this attribute?

leSamo:
@gkarat I wasn't able to put it in root level so I just now put it in each main section. Shallow test is not needed, this is not a permanent solution just an experiment to see how intrusive it would be to have the whole app suppressed in Hotjar recordings.

PR #654, opened 2023-08-10
fix(RHINENG-1032): create baseline navigates to undefined

johnsonm325:
Repro:

Navigate to Baselines
Click Create Baseline
Create baseline from scratch
Enter baseline name and click Create Baseline
The result in stage will be Error 400: malformed UUIDs (see screenshots).  However, the baseline does get created.

This is a react 18 bug and you won't be able to easily reproduce it in dev. It was an issue with a navigate trying to run before an api request had time to finish. It's part of the new feature with bulk requests.

You'll have to review the code, push to stage and check there to make sure it works.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/src/SmartComponents/BaselinesPage/CreateBaselineModal/CreateBaselineModal.js
+++ b/src/SmartComponents/BaselinesPage/CreateBaselineModal/CreateBaselineModal.js
@@ -1,6 +1,6 @@
-import React, { Component } from 'react';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
-import { connect } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 import { Button, Modal, Radio, TextInput, Form, FormGroup, ValidatedOptions } from '@patternfly/react-core';
 import { sortable, cellWidth } from '@patternfly/react-table';
 import { addNewListener } from '../../../store';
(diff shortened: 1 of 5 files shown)


src/SmartComponents/BaselinesPage/CreateBaselineModal/CreateBaselineModal.js
```diff
@@ -142,7 +148,6 @@ export class CreateBaselineModal extends Component {
                     await createBaseline(newBaselineObject);
```
bastilian, line 148: There may be more to this. 

`createBaseline` is not an async function, it is a redux action, but here it is not really dispatched, it is just called, so redux is never involved in this.

The proper fix should involve removing the `await`s here and properly dispatch the redux `createBaseline` action. 

The fix should also not involve `componentsDidUpdate`.

johnsonm325: @bastilian It is dispatched here https://github.com/RedHatInsights/drift-frontend/blob/3b54a28b5a267de1901acb9747c4022f93cd4968/src/SmartComponents/BaselinesPage/CreateBaselineModal/CreateBaselineModal.js#L430

Also, with React 18, await doesn't act the same way. The navigate gets called before the `await createBaseline` gets finished so it navigates to `undefined`. Using `componentDidUpdate` to wait for the baselineData to come through is the best way to ensure that the `createBaseline` completed.

Fewwy (approved):
lgtm

PR #665, opened 2023-09-06
feat(ESSNTL-3722): Add groups column to baselines detail page

adonispuente:
This PR implements the group filtering in Drift baselines table page. Currently the backend PR associated with this is only on QA, but as for the testing of the PR, simply navigate to a baseline's details, and then select the groups filter.


gkarat (approved):
LGTM! Thanks @adonispuente 

code-wise: we will improve the business logic here for fetching IDs together with refactoring. in this PR's implementation it fulfils the feature requirements 👍🏼

PR #673, opened 2023-10-13
refactor(RHINENG-1608): Make setHistory a hook

adonispuente:
There should be no change in behavior. The goal of this is to pass in the function as a hook , and hopefully that resolves the issue in https://issues.redhat.com/browse/RHINENG-1608 

There is no way to reproduce the bug in dev environments, so this review would just be ensuring there are no behavioral changes locally when running.


bastilian:
@adonispuente is this really ready for review? It seems to just rename the function, but it does not really turn it into a hook, yet.

adonispuente:
@bastilian Sorry about that, I hadnt sent up the most updated commit. I ended up finding some issues in that version, this is the simplest version of the hook that doesnt result in needing to touch other things.

PR #686, opened 2024-01-12
[RHINENG-1025] update class components to fix table

johnsonm325:
This PR updates a number of class components to functional components and further breaks up the drift table into more manageable components. Functionality should remain the same. The hope is that this will fix the column resizing issue on the drift table in stage/prod.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


src/SmartComponents/DriftPage/DriftToolbar/__tests__/DriftToolbar.tests.js
```diff
@@ -1,11 +1,15 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { Provider } from 'react-redux';
+import { shallow, mount } from 'enzyme';
```
mkholjuraev, line 3: RTL should be used instead of enzyme.

mkholjuraev (commented):
Tests that assert on the functional calls are not encouraged in RTL as they usually test implementation details rather than user behavior on the UI. I see that most of the new tests in this PR are testing implementation detail. Would you say we should also opt for tests on user behavior while we are on this? Reference: https://medium.com/@ankitsingh80741/a-guide-to-effective-testing-with-react-testing-library-475a3bf1988c

src/SmartComponents/DriftPage/DriftToolbar/__tests__/DriftToolbar.tests.js
johnsonm325: yes, but I didn't want to take time to convert this entire test file into rtl. that can be done later. I just did this to get a better snapshot temporarily.

adonispuente (approved):
Muslimjon has some comments to be addressed, but overall when running locally, no functionality seems to be affected. I wasnt able to encounter the issue locally (but I believe thats one of the main issues with it). 
Overall the refactor is great. Theres a point to be made with converting test but it doesnt all have to be done with this specific PR.

PR #690, opened 2024-03-22
fix(RHINENG-8631): upgrade axios to ^1.6.8 version

mkholjuraev:
This fixes the RHINENG-8631 by updating the axios version to ^1.6.8


gkarat (commented):
I believe the same `overrides` thingy applies here as I mentioned in https://github.com/RedHatInsights/compliance-frontend/pull/2045.

gkarat (approved):

PR #698, opened 2024-05-31
fix(RHINENG-5473): Do not show CentOS systems in modals

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-5473.

This makes sure no modals show CentOS systems nor allowing users to filter out the CentOS systems. Drift only supports RHEL systems at the moment.

## How to test

1. Use an account with at least one CentOS system registered in Inventory
2. Go to /insights/drift and verify that you are not able to add any CentOS system to comparison
3. Create or navigate to any baselines, switch to Associated systems
4. Make sure you are the same way not able to associate any CentOS system with this baseline.

### DIFF
--- a/config/setupTests.js
+++ b/config/setupTests.js
@@ -35,7 +35,7 @@ jest.mock('react-router-dom', () => ({
 }));
 
 jest.mock('@redhat-cloud-services/frontend-components/Inventory', () => ({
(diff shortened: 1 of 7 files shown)


mkholjuraev (approved):
LGTM! The filter is passed to the API properly on both pages.

